PR #2, opened 2018-05-29
Splash page

alocke12992:
Redesigns the splash page and resizes the video component. @frehner


website/static/css/custom.css
```diff
@@ -1,11 +1,34 @@
 /* your custom css */
+@import url('https://fonts.googleapis.com/css?family=Montserrat:300,400|Roboto:100,400');
```
frehner, line 2: can you use the `stylesheets` option in `siteConfig.js` for this? I don't know for sure, so just asking. 

https://docusaurus.io/docs/en/site-config.html#optional-fields

website/static/css/custom.css
```diff
@@ -1,11 +1,34 @@
...
 }
 
+.homeContainer{
+  background: url('../img/splash.svg');
+  background-repeat: no-repeat;
+  background-position: 85% 100%; 
+}
+
+.projectTitle {
+  font-family: 'Roboto', sans-serif;
```
frehner, line 17: I think you can use `$mainFont` here, right? see the `fonts` section of this https://docusaurus.io/docs/en/site-config.html#optional-fields

I could be wrong though.

website/static/css/custom.css
```diff
@@ -1,11 +1,34 @@
...
+  font-weight: 100 !important;
+  font-size: 500%;
+  letter-spacing: .9rem;
+}
+
+.subHeader {
+  font-family: 'Montserrat', sans-serif;
```
frehner, line 24: same here; should be able to use `$accentFont` right?

website/static/css/custom.css
```diff
@@ -1,11 +1,34 @@
...
+  font-weight: 100 !important;
+  font-size: 110%;
+  letter-spacing: .2rem;
+  color: #889aaa;
```
frehner, line 28: could also setup your color palette in `siteConfig.js` and use the variables for those here -- big advantage is that you can more easily change the color once and it's updated globally. I think?

https://docusaurus.io/docs/en/site-config.html#optional-fields under `colors`

website/static/img/beamery.svg
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg width="300px" height="255px" viewBox="0 0 300 255" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <!-- Generator: Sketch 3.7.2 (28276) - http://www.bohemiancoding.com/sketch -->
+    <title>beamery-vertical-secondary</title>
```
frehner, line 4: how much do we care about accessibility? (probably not as much as we should, to be honest).

if we care a bit, then I think updating the `<title>` here and adding them on the other SVGs would be good....?

website/static/css/custom.css
alocke12992: They say that should work, but I found when trying that it would reject the variable and revert to browser defaults.

frehner: huh strange. I think I had it working in some places -- maybe we can work together and figure out what's going on?

website/static/img/beamery.svg
alocke12992: great point! I hadn't even thought about that.

website/static/css/custom.css
alocke12992: After some playing around with it I was able to figure it out!

frehner: glad to hear! what was the issue?

PR #4, opened 2018-06-01
Contributors

alocke12992:


website/static/js/index.js
```diff
@@ -0,0 +1,24 @@
...
+            <div class='contributor'>
+              <a href='${item.html_url}' target='_blank' ref='noopener noreferrer'>
+                <h2 class='contributorLogin'>
+                  ${item.login}
+                </h2>
+              </a>
+              <img class='contributorImg' src=${item.avatar_url} alt='${item.login} avatar' />
+            </div>  
+            `
+          )).join('')}
```
frehner, line 20: want to have a `.catch` here just in case?

website/pages/en/contributors.js
```diff
@@ -0,0 +1,22 @@
+const React = require('react')
+const CompLibrary = require('../../core/CompLibrary.js'); 
+const Container = CompLibrary.Container;
+const GridBlock = CompLibrary.GridBlock;
```
kentmclean: Unused item?

website/static/js/index.js
alocke12992: smart. Thanks for the .catch ;)

website/pages/en/contributors.js
alocke12992: Yup, good catch.

PR #16, opened 2018-08-03
inconsistency with single-spa-config file name

benschinn:
On the [Start From Scratch](https://single-spa.js.org/docs/starting-from-scratch.html) page, the _single-spa-config_ file is referenced at times as `single-spa-config.js` and at times `single-spa.config.js`. 
I am not sure how important the naming convention is as far as whether it should be `single-spa-config.js` or `single-spa.config.js`, but as I was following the documentation, I was able to get the example easily working with having the file named `single-spa.config.js`, but ran into issues when the file was named `single-spa-config.js`.

### DIFF
--- a/docs/starting-from-scratch.md
+++ b/docs/starting-from-scratch.md
@@ -133,7 +133,7 @@ module.exports = {
     },
     node: {
         fs: 'empty'
-    }, 
+    },
     resolve: {
         modules: [
             __dirname,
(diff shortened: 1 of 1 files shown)


frehner (approved):
Ah, good catch. It almost appears to me like there was some confusion between the actual file name and the documentation's markdown file name here https://github.com/CanopyTax/single-spa.js.org/blob/master/docs/single-spa-config.md

AFAIK there isn't a convention or best practice as far as the name of the file goes.

So thank you for finding and fixing these!

PR #17, opened 2018-08-09
How to make Vue router work with single-spa

factoidforrest:
None of single-spa's examples or documentation attempt to use Vue's router.  If they did, users would know that in order to use the router with single-spa, it needs to be in "history" mode so as not to conflict with single-spa's hash based routing.

### DIFF
--- a/docs/ecosystem-vue.md
+++ b/docs/ecosystem-vue.md
@@ -39,6 +39,15 @@ export const unmount = [
 ];
 ```
 
+## Routing
+In order for Vue router to work with Single-Spa's routing, it needs to be in `history` mode, not in `#` hash mode which is the default.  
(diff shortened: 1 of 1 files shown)


factoidforrest:
Actually I'm not sure this is right either.  What is the official way to do vue routing in single-spa?

internettrans:
Apologies for the slow response here -- your normal vue router should work seamlessly with single-spa as long as the following things are met:

1) The vue router needs to remove event listeners for `hashchange` and `popstate` whenever the vue app is unmounted by single-spa
2) The vue routes need to be sub routes of the route specified in the vue app's [activity function](https://single-spa.js.org/docs/configuration.html#activity-function)

PR #22, opened 2018-10-12
Lock down babel-loader to version 7

jonaslu:
The latest babel-loader (v8.x) depends on babel v7.x which has breaking changes with the rest of the dependencies in the example. Lock down babel-loader version to 7.x to keep the example working.


frehner (approved):
thanks!

PR #23, opened 2018-10-13
Correct the bower package name in bower

jonaslu:
It's single-spa-ember, not single-spa-examples, correct it in the spa ember example.


frehner (approved):
Thanks

PR #27, opened 2018-11-28
fix twitter url in footer

kartikag01:
Twitter Link on the footer of https://single-spa.js.org/ was leading to https://twitter.com/ which was supposed to be https://twitter.com/Single_spa/
Fixed twitter link in footer code.


frehner:
Thanks!

PR #36, opened 2019-01-30
Adding Jump Software as a single-spa user

keithhalterman:


keithhalterman:
Cool, Looks like @joeldenning already added this.

PR #41, opened 2019-02-20
Updated link to angular examples repo

StephenRansom:
Previously linked to ( https://github.com/CanopyTax/single-spa-examples/blob/master/src/angular2/angular2.app.js ) which gives a 404


frehner (approved):
Thank you for your contribution!

PR #48, opened 2019-03-25
Docs/upgrade migration guides

filoxo:
I wasn't intending to make so many changes, but... I kind of got carried away. 

About these changes: 
- mostly simplifies the wording in some places
- fixes/makes consistent the formatting
- Add function signatures to the APIs page
    - I think this style helps define the code much better, but I'm open to feedback
- made the comparison table in "Separating Applications" more accessible
    - but when I was testing, I found that for some reason VoiceOver wasn't associating the column/row headers correctly. Not sure if this has something to do with markdown rendering.
- Adds some utility style classes like `.deprecated` and `.sr-text` to use throughout


filoxo:
Preview here: https://sspa-docs-ddkntfjqbr.now.sh

docs/api.md
```diff
@@ -4,211 +4,411 @@ title: single-spa API
...
+`undefined`
 
 ## declareChildApplication
-`declareChildApplication` is deprecated use `registerApplication` instead
+
+<p class="deprecated">Deprecated</p>
+
+Use [`registerApplication`](#registerapplication) instead.
 
 ## start
```
frehner, line 47: this all looks amazing! however, I noticed that the right sidebar isn't picking up on the headers anymore on your example repo. Wonder what changed?

frehner (commented):
One small thing I noticed. Otherwise things look pretty nice. 

However, it's a big review so I didn't look it all over at this time.

docs/api.md
filoxo: Oh weird, I hadn't caught that. I'll look into why.

docs/api.md
```diff
@@ -4,211 +4,411 @@ title: single-spa API
...
-This event will get fired after each hashchange, popstate, or triggerAppChange, even if no changes
-to registered applications were necessary. Sample usage of this event might look like this:
+Will create and mount a [single-spa parcel](parcels-overview.md).
+
+> Note: parcels do not automatically unmount. Unmounting will need to be triggered manually.
+
+<h3>arguments</h3>
+
+<dl class="args-list">
+	<dt>parcelConfig: ???</dt>
```
filoxo, line 274: Oops forgot to remove this. I plan on submitting updated parcels docs in a future PR.

docs/api.md
filoxo: Turns out it was this HTML comment here that was borking things. https://github.com/CanopyTax/single-spa.js.org/pull/48/files#diff-0f426376b964e609c5c4618eb8bed3b2R18

I'll remove it since it wasn't adding anything, just was an attempt at another style of displaying the args/returns.

frehner:
This is amazing work, and I like how it looks so far. Let's go for it and if there are any problems we can just do followup PRs -- I don't want you to have to keep this PR up to date with new changes coming in.

PR #52, opened 2019-04-25
Fix single-spa users images links

dzienisz:
Hey 👋

I found and fixed small bug with links to single-spa users images.

I would appreciate if you could merge my first contribution to single-spa.

Thanks


frehner:
Thanks for your contribution! Were these images broken on a certain page? They appear to be working on the homepage and https://single-spa.js.org/users.html but honestly I don't know if/where any other places use that array.

dzienisz:
hmm that's funny, they were broken for me yesterday 😅 maybe that was a bug in my browser

frehner:
No worries! Thanks for taking a look.

PR #65, opened 2019-06-23
Updating lifecycle props.

internettrans:
See https://github.com/CanopyTax/single-spa/blob/master/src/lifecycles/prop.helpers.js#L4


TheMcMurder (approved):
🐙

PR #67, opened 2019-07-05
Switching errors to use appOrParcelName property. See canopytax/single-spa#324.

internettrans:
See canopytax/single-spa#324.


frehner (approved):
🐽

PR #69, opened 2019-07-08
Adding CRA to FAQ

TheMcMurder:


docs/faq.md
```diff
@@ -85,3 +85,6 @@ With the [recommended setup](#is-there-a-recommended-setup), the process general
...
 * Server render your `index.html` with the import map inlined. This does not mean that your DOM elements need to all be server rendered, but just the `<script type="systemjs-importmap>` element. Provide an API that either updates a database table or a file local to the server.
 * Have your import map itself on a CDN, and use [import-map-deployer](https://github.com/CanopyTax/import-map-deployer) or similar to update the import map during your CI process. This method has a small impact on performance, but is generally easier to setup if you don't have a server-rendered setup already. (You can also [preload](https://developer.mozilla.org/en-US/docs/Web/HTML/Preloading_content) the import map file to help provide a small speed boost). See [example travis.yml](https://github.com/openmrs/openmrs-esm-root-config/blob/master/.travis.yml). Other CI tools work, too.
+
+## Create react app
+Currently create react app requires ejecting or using a tool to modify the webpack config like: https://github.com/timarney/react-app-rewired/blob/master/README.md consider some of the alternatives for working with single-spa: https://github.com/facebook/create-react-app#popular-alternatives
```
frehner:
```suggestion
Currently Create React App (CRA) requires [ejecting](https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/template/README.md#npm-run-eject) or [using a tool](https://github.com/timarney/react-app-rewired/blob/master/README.md) to modify the webpack config.  You can also consider some of the [popular alternatives to CRA](https://github.com/facebook/create-react-app#popular-alternatives).

The reason for this is because, when you use the [recommended setup](#is-there-a-recommended-setup), your webpack's [`output.libraryTarget`](https://webpack.js.org/configuration/output/#outputlibrarytarget) needs to be either `System`, `UMD`, or `AMD`.  CRA does not allow you to output to these formats without ejecting or using another tool. 
```

frehner (commented):
thanks! cleaned up the text a little bit, but that should be it. :)

internettrans (commented):
If you wanted to take the docs to the next level, you could specify the high level things that need to change from the CRA webpack config:

1) get rid of the webpack optimizations block because they add multiple webpack chunks that don’t load each other
2) get rid of html webpack plugin
3) change output library target to “umd” or “system”

frehner (approved):
🐽

frehner:
@joeldenning oh dang just saw your comment. Yeah we should add that in a followup MR

PR #92, opened 2019-08-12
Adding documentation for custom props in Vue.

internettrans:
See https://github.com/CanopyTax/single-spa-vue/issues/19 for the issue that made me realize we had no documentation for custom props and vue.


frehner (approved):
🐽

PR #105, opened 2019-10-09
Simplifying introductory statement

TheMcMurder:
Hopefully this makes parcels _slightly_ less confusing

### DIFF
--- a/docs/parcels-overview.md
+++ b/docs/parcels-overview.md
@@ -3,9 +3,7 @@ id: parcels-overview
 title: Parcels
 sidebar_label: Overview
 ---
-
-A single-spa parcel is a bundle of functionality (like an application or component) meant to be mounted manually
-by an application. Parcels use similar methodology as applications but are mounted by a manual function call rather than the activity function.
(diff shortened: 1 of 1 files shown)


docs/parcels-overview.md
```diff
@@ -3,9 +3,7 @@ id: parcels-overview
 title: Parcels
 sidebar_label: Overview
 ---
-
-A single-spa parcel is a bundle of functionality (like an application or component) meant to be mounted manually
-by an application. Parcels use similar methodology as applications but are mounted by a manual function call rather than the activity function.
+A single-spa parcel can best be thought of as a single-spa style custom element. It is a bundle of functionality (like a component) meant to be mounted manually by an application. Parcels use similar methodology as registered applications but are mounted by a manual function call rather than the activity function.
```
internettrans: I've added a suggestion for the wording that removes the comparison to custom elements. If you do want to keep the comparison intact, I'd recommend calling them web components because people know that term better. Additionally, the phrase `single-spa style custom element` needs a hyphen: `single-spa-style` in order for them to know that the style applies to single-spa instead of to custom element. Even easier would be just to remove the word `style` altogether.

```suggestion
A single-spa parcel is a framework agnostic component. It is a chunk of functionality meant to be mounted manually by an application, without having to worry about which framework was used to implement the parcel or application. Parcels use similar methodology as registered applications but are mounted by a manual function call rather than the activity function.
```

PR #107, opened 2019-10-10
Update parcels-overview.md

TheMcMurder:
Adding more parcels documentation

I don't think we address this direct enough in the current documentation, so I wanted to add a section on it.

### DIFF
--- a/docs/parcels-overview.md
+++ b/docs/parcels-overview.md
@@ -185,3 +185,33 @@ componentDidMount() {
 }
 ...
 ```
+## `mountRootParcel` vs `mountParcel`
+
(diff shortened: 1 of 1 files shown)


TheMcMurder:
I also wanted to add a warning about parcels as an advanced feature of single spa but IIRC others have previously expressed dissent to that idea so I left it out.

If everyone else agrees that parcels are "advanced" then I'll add something like that in another PR. If there is still disagreement I think these changes will help with confusion about `mountParcel` and `mountRootParcel`

filoxo (commented):
This really really clarified the need for the React context with Parcels.

filoxo (approved):

PR #118, opened 2019-11-08
Fix redirects 2

lex111:
:man_facepalming:


frehner:
whoops my bad, I should have looked at it more closely

PR #119, opened 2019-11-08
Update index.js

frehner:
handle situation where you have the path `/docs` from adding two `//` e.g. `/docs//`


frehner:
@lex111 does this look ok?

frehner:
actually I'm not sure that's the right solution. one second

to reproduce, go to this url https://single-spa.js.org/docs

PR #120, opened 2019-11-08
Fix redirects

lex111:
Demo build - https://build-jb3zykk1v.now.

Finally it should work correctly, at least there are no problems on the demo website, please see.

cc @frehner

### DIFF
--- a/website/src/plugins/docusaurus-plugin-redirects/src/index.js
+++ b/website/src/plugins/docusaurus-plugin-redirects/src/index.js
@@ -1,5 +1,6 @@
 const path = require('path');
 const fs = require('fs');
+const normalizeUrl = require('@docusaurus/utils').normalizeUrl;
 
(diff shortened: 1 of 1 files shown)


frehner:
code looks good to me

demo site is dead.

lex111:
Correct link - https://build-jb3zykk1v.now.sh/

frehner:
thanks again

lex111:
@frehner unfortunately, the issue has not been resolved (yes, yes, it also annoys me already!). I think the matter is in the hosting provider, because now the redirects work fine in Now Zeit, and something goes wrong on the GitHub Pages.
But in fact, you correctly proposed a solution in https://github.com/CanopyTax/single-spa.js.org/pull/119/files
I can’t come up with the best yet, but yours works at 100%.

lex111:
https://github.com/CanopyTax/single-spa.js.org/pull/121/files

PR #128, opened 2019-12-02
Add 'security' content to FAQ

filoxo:


docs/faq.md
```diff
@@ -110,3 +110,18 @@ Single spa supports code splits. There are so many ways to code split we won't b
...
+Outside of that, web applications may use the following resources that have their own security considerations that you may need to become familiar with.
+
+- [ES6 module dynamic imports](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import)
+  - Webpack-based applications use [Webpack's implementation of dynamic imports](https://webpack.js.org/guides/code-splitting/#dynamic-imports)
+- [Cross-Origin Resource Sharing (CORS)](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS)
+- [Content Security Policy (CSP)](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP)
+  - module imports specifically relate to [CSP `script-src`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src)
+- Import-maps are also governed by CSP
+  - See also ["Supplying out-of-band metadata for each module"](https://github.com/WICG/import-maps/blob/master/README.md#supplying-out-of-band-metadata-for-each-module)
+  - See also [import-maps script “integrity” attribute](https://github.com/WICG/import-maps/issues/174)
```
internettrans: Subresource integrity is not a part of CSP and shouldn’t be indented under that section. This sentence makes it sound like SRI is a topic that applies specifically to import maps, even though they’re not at all.

I think SRI should be its own bullet point with a sub bullet point linking to SRI for import maps

docs/faq.md
```diff
@@ -110,3 +110,18 @@ Single spa supports code splits. There are so many ways to code split we won't b
 
     * For SystemJS 2-5: Find a code example [here](https://gitlab.com/TheMcMurder/single-spa-portal-example/blob/master/people/src/set-public-path.js#L3)
 1. Set either [`output.jsonpFunction`](https://webpack.js.org/configuration/output/#outputjsonpfunction) or [`output.library`](https://webpack.js.org/configuration/output/#outputlibrary) to ensure that each app's webpack doesn't collide with other apps' webpack. `jsonpFunction` is preferred.
+
+## Does single-spa require additional security considerations?
+
+No. single-spa does not add, deviate, or attempt to bypass any browser JavaScript security measures. The security needs of your applications are the same as if you did not use single-spa.
+
+Outside of that, web applications may use the following resources that have their own security considerations that you may need to become familiar with.
```
internettrans: Maybe end this line with a colon instead of period

docs/faq.md
```diff
@@ -110,3 +110,18 @@ Single spa supports code splits. There are so many ways to code split we won't b
...
+
+- [ES6 module dynamic imports](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import)
```
internettrans, line 120: The links here don’t really explain security as much as just the topics. Might be nice to link to both the topic and a discussion about its security. If those links are hard to find I’m okay with omitting them, though

filoxo: Yea, I couldn't find anything that delved into security specifically with each of these topics, but I figured these links would at least be where you'd start learning about how they work. 

Maybe this would be a good link for module dynamic imports security? https://github.com/w3c/webappsec-csp/issues/243

docs/faq.md
filoxo: Sounds good.

docs/faq.md
filoxo: I updated some of the links to point to the sections that talk about security specifically.

internettrans (approved):

PR #135, opened 2019-12-10
Second batch of error code pages

internettrans:
This adds six more documentation pages for single-spa minified errors (in v5)


website/src/components/Errors/Codes/2.js
```diff
@@ -0,0 +1,45 @@
+import React from 'react'
+import CodeSnippet from '../CodeSnippet'
+
+export default function ErrorCode2() {
```
internettrans, line 4: See https://github.com/CanopyTax/single-spa/blob/7dcbe87c69d1e0e1e74517b1a9b7c279edb8be93/src/parcels/mount-parcel.js#L34

This PR covers most of the error messages within that file.

frehner (approved):

PR #143, opened 2019-12-30
Adding pages for recommended setup and microfrontends concept

internettrans:
These pages are in the "next" version in docusaurus, which means you can only get to them by manually typing in the URL. I plan to release them as part of the single-spa@5 release.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -11,13 +11,7 @@ The code for a route is called an "application", and each can (optionally) be in
 The applications can all be written in the same framework, or they can be implemented in different frameworks.
 
 ## Is there a recommended setup?
-We recommend a setup that uses in-browser ES modules + [import maps](#what-are-import-maps) (or [SystemJS](https://github.com/systemjs/systemjs) to polyfill these if you need better browser support).  This setup has several advantages:
-
-1. Common libraries are easy to manage, and are only downloaded once. If you're using SystemJS, you can also [preload them](https://developer.mozilla.org/en-US/docs/Web/HTML/Preloading_content) for a speed boost as well.
-1. Sharing code / functions / variables is as easy as `import/export`, just like in a monolithic setup
-1. Lazy loading applications is easy, which enables you to speed up initial load times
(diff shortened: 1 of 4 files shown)


docs/faq.md
```diff
@@ -11,13 +11,7 @@ The code for a route is called an "application", and each can (optionally) be in
...
-
-1. Common libraries are easy to manage, and are only downloaded once. If you're using SystemJS, you can also [preload them](https://developer.mozilla.org/en-US/docs/Web/HTML/Preloading_content) for a speed boost as well.
-1. Sharing code / functions / variables is as easy as `import/export`, just like in a monolithic setup
-1. Lazy loading applications is easy, which enables you to speed up initial load times
-1. Each application (AKA microservice, AKA ES module) can be independently developed and deployed. Teams are enabled to work at their own speed, experiment (within reason as defined by the organization), QA, and deploy on thier own schedules. This usually also means that release cycles can be decreased to days instead of weeks or months
-1. A great developer experience (DX): go to your dev environment and add an import map that points the application's url to your localhost. See "[What is the DX like?](#what-is-the-developer-experience-dx-like)" for more details.
```
TheMcMurder, line 20: Was this all deleted? I don't see it in the recommended setup page.

docs/recommended-setup.md
```diff
@@ -0,0 +1,37 @@
+---
+id: recommended-setup
+title: The Recommended Setup
+sidebar_label: Overview
+---
+
+Single-spa itself is not opinionated about your build tools, CI process, or local development workflow. However, to implement single-spa you will have to figure all of those things out (and more). To help you decide how to approach these problems, the single-spa core team has put together a "recommended setup" that gives an opinionated approach to solving the practical problems of microfrontends.
+
```
TheMcMurder, line 8: I love the direction this is going. I'd still like to keep the text summary from before as the headline content here with the videos being supplemental

internettrans:
Updated with your feedback

TheMcMurder (approved):

frehner:
Agreed with Justin above - I love the video content but video is a lot harder to consume for many people over text. It would be great if we had both - even if it’s a transcript of the video or something

internettrans:
I agree - we should have both written content and the video. @filoxo showed interest in helping write the content.

PR #144, opened 2020-01-06
docs: Adding return vue instance in mount method(ecosystem-vue.md)

kwen8:
update docs

### DIFF
--- a/docs/ecosystem-vue.md
+++ b/docs/ecosystem-vue.md
@@ -7,13 +7,17 @@ sidebar_label: Vue
 single-spa-vue is a helper library that helps implement [single-spa registered application](configuration#registering-applications) [lifecycle functions](building-applications.md#registered-application-lifecycle) (bootstrap, mount and unmount) for for use with [Vue.js](https://vuejs.org/). Check out the [single-spa-vue github](https://github.com/CanopyTax/single-spa-vue).
 
 ## Starter repo
+
 For a full example, see [coexisting-vue-microfrontends](https://github.com/joeldenning/coexisting-vue-microfrontends).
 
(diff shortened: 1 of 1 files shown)


docs/ecosystem-vue.md
```diff
@@ -7,45 +7,51 @@ sidebar_label: Vue
...
 The CLI Plugin does the following for you:
-1) Modify your webpack config so that your project works as a single-spa application or parcel.
-2) Install [single-spa-vue](https://github.com/CanopyTax/single-spa-vue).
-3) Modify your `main.js` or `main.ts` file so that your project works as a single-spa application or parcel.
-4) Add a `set-public-path.js` that will use `systemjs-webpack-interop` in order to set the public path of your application.
+
+1. Modify your webpack config so that your project works as a single-spa application or parcel.
+2. Install [single-spa-vue](https://github.com/CanopyTax/single-spa-vue).
+3. Modify your `main.js` or `main.ts` file so that your project works as a single-spa application or parcel.
+4. Add a `set-public-path.js` that will use `systemjs-webpack-interop` in order to set the public path of your application.
```
frehner, line 32: can we change these to just `1.` for all of them? that way we don't need to keep the list ordered in future revisions. :)

frehner (commented):
small comment, otherwise looks good. thanks!

docs/ecosystem-vue.md
kwen8: Sure, I think it's ok

internettrans (approved):
I think it's fine without changing all to 1's. Better to merge this in than let it sit waiting on that I think

PR #147, opened 2020-01-14
Clarifying how to specify which dom element applications are mounted to

internettrans:
See related https://github.com/CanopyTax/single-spa-vue/issues/31


frehner (approved):
🐽

PR #148, opened 2020-01-14
Adding six more error code documentation pages.

internettrans:
See single-spa@5's error codes for more details.

### DIFF
--- a/website/src/components/Errors/Codes/25.js
+++ b/website/src/components/Errors/Codes/25.js
@@ -1,0 +1,23 @@
+import React from 'react'
+import CodeSnippet from '../CodeSnippet'
+
+export default function ErrorCode25(props) {
+  return (
+    <>
+      <h1>#25: Cannot unregister application - does not exist</h1>
+      <p>
+        You cannot unregister application {props.getErrorCodeArg(0)} because no application with that name was registered.
+      </p>
+      <h2>To fix:</h2>
+      <div>
+        <CodeSnippet>
+          {`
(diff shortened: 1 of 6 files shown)


website/src/components/Errors/Codes/25.js
```diff
@@ -0,0 +1,23 @@
...
+    <>
+      <h1>#25: Cannot unregister application - does not exist</h1>
+      <p>
+        You cannot unregister application {props.getErrorCodeArg(0)} because no application with that name was registered.
+      </p>
+      <h2>To fix:</h2>
+      <div>
+        <CodeSnippet>
+          {`
+            __SINGLE_SPA_DEVTOOLS__.exposedMethods.unregisterApplication(
```
frehner, line 15: wait this is the fix to this error? not sure I understand

internettrans: Error code 25 is about the unregisterApplication API, which is only used by the devtools. So the entire error code is sorta pointless. With each error code I've been trying to provide a code example that shows how to fix the problem, because I think that self contained code samples are the most helpful thing.

The real fix is just to make sure you call unregisterApplication with the name of an actually registered application. But I couldn't think of a good way to show that so I just showed how to call unregisterApplication at all.

What are your thoughts on what would be most helpful?

frehner: ah, if it's an error that only happens with devtools, then that's ok - no issues from me.

frehner (approved):
🐽

PR #150, opened 2020-01-16
Updating user showcase

internettrans:


website/src/data/users.js
```diff
@@ -17,12 +17,6 @@ const users = [
     infoLink: 'https://beamery.com/',
     pinned: true,
   },
-  {
-    caption: 'Jump',
```
internettrans, line 21: rip

frehner: technically you could replace it with `Nuvi` but we'd probably have to get their consent, and that's not something I can do at this moment :D

frehner (approved):

website/src/data/users.js
internettrans: Let's ask the CEO via a Glassdoor DM? 😄

PR #152, opened 2020-01-21
Adds single-spa example of Option 2: NPM packages (v4)

jualoppaz:
Adds a single-spa example that shows how to setup Option 2: NPM packages strategy for splitting applications

### DIFF
--- a/website/versioned_docs/version-4.x/separating-applications.md
+++ b/website/versioned_docs/version-4.x/separating-applications.md
@@ -132,7 +132,13 @@ Changing which javascript file is loaded for each child application can be done 
           </li>
         </ul>
       </td>
-      <td>Contributions accepted!</td>
+      <td>
+        <ul>
+          <li>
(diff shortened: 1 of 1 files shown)


internettrans (changes requested):
👍 thanks. Can you also add this to the v5 docs (they are pending release). The file to modify is https://github.com/single-spa/single-spa.js.org/blob/master/docs/separating-applications.md. We will be releasing v5 relatively soon

jualoppaz:
@joeldenning here you have: https://github.com/single-spa/single-spa.js.org/pull/153 . I tried to edit it directly from docs page and I supposed that I was editing master instead of v4 docs.

This is very minimalist example. I'll improve it along next week

internettrans (approved):
👍 thanks. Your work creating the example is appreciated.

jualoppaz:
@joeldenning thanks to you. We are starting using single-spa in my job and this dedication has been very good for me 😉

PR #159, opened 2020-01-26
Adds note about workaround for Angular Router (v4)

jualoppaz:


internettrans:
Like I said in #160, I don't think this information is specific to Angular 4/5 and should just be part of the regular docs instead of specific to a version

jualoppaz:
V4 refers to single-spa docs version and not to Angular Router. Sorry for the confusion.

internettrans (approved):
Ah my mistake - thanks for this. Btw you can modify more than one file at a time if you fork and clone this repo, then push up the code from your computer.

jualoppaz:
Ok @joeldenning . I'll have it in mind this next time. Thanks.

PR #160, opened 2020-01-26
Adds note about workaround for Angular Router (v5)

jualoppaz:


internettrans:
Is this information specific to angular 5? I think this applies to all versions of angular.

jualoppaz:
V5 refers to next single-spa docs version and not to Angular Router. Sorry for the confusion.

internettrans (approved):

PR #163, opened 2020-02-11
Add FAQ about frameworks in the single-spa-config

TheMcMurder:
I think this is a common point of confusion and I've noticed an increase in questions about frameworks in the root config.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -13,6 +13,9 @@ The applications can all be written in the same framework, or they can be implem
 ## Is there a recommended setup?
 Yes, here is [the documentation for our recommended setup](/docs/recommended-setup/).
 
+## Should I put frameworks in the single-spa-config/root-application?
+No. We strongly encourage that your single-spa-config or root application does not use any javascript ui-frameworks (React, Angular, Angularjs, Vue, etc). In our experience a plain javascript module is best for the single-spa-config and only the registered applications actually use ui-frameworks (angular, react, vue, etc). 
+
 ## What is the impact to performance?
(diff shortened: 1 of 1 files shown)


docs/faq.md
```diff
@@ -13,6 +13,9 @@ The applications can all be written in the same framework, or they can be implem
 ## Is there a recommended setup?
 Yes, here is [the documentation for our recommended setup](/docs/recommended-setup/).
 
+## Should I put frameworks in the single-spa-config/root-application?
```
frehner:
```suggestion
## Should I have a parent/root app and children apps?
```

docs/faq.md
```diff
@@ -13,6 +13,9 @@ The applications can all be written in the same framework, or they can be implem
...
+No. We strongly encourage that your single-spa-config or root application does not use any javascript ui-frameworks (React, Angular, Angularjs, Vue, etc). In our experience a plain javascript module is best for the single-spa-config and only the registered applications actually use ui-frameworks (angular, react, vue, etc).
```
frehner, line 17:
```suggestion
No. We strongly encourage that your single-spa-config or root application does not use any javascript ui-frameworks (React, Angular, Angularjs, Vue, etc). In our experience a plain javascript module is best for the single-spa-config and only the registered applications actually use ui-frameworks (angular, react, vue, etc). 

Why? You end up creating a structure that has all the disadvantages of microservices without any of the advantages: your apps are now coupled together and you have to change multiple apps at the same time in order to make updates. Good microservices are completely **independent**, and this pattern breaks that.
```

frehner (commented):
feel free to take/throw out my suggestions here.

frehner (approved):
🐽

internettrans:
Oo yes I completely agree

PR #169, opened 2020-02-14
Update parcels-overview.md

TheMcMurder:


internettrans (approved):
This will only change single-spa@5 docs - if you want to change v4 too that's a separate file. But this looks good to me.

PR #171, opened 2020-02-17
Adding documentation for single-spa-dojo. See dojo.io

internettrans:
See https://github.com/single-spa/single-spa-dojo

### DIFF
--- a/docs/ecosystem-dojo.md
+++ b/docs/ecosystem-dojo.md
@@ -1,0 +1,67 @@
+---
+id: ecosystem-dojo
+title: single-spa-dojo
+sidebar_label: Dojo
+---
+
+[![Build Status](https://travis-ci.com/single-spa/single-spa-dojo.svg?branch=master)](https://travis-ci.com/single-spa/single-spa-dojo)
+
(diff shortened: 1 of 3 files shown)


filoxo:
This is definitely only just a rudimentary understanding but (if) [single-spa-dojo doesn't actually unmount from the dom](https://github.com/single-spa/single-spa-dojo/pull/4/files#diff-d952140975bd9283bae58ac837c1d557R69), should that be documented here as well? Otherwise LGTM

internettrans:
It does actually unmount from the DOM. But it doesn't clean up all listeners yet. See https://github.com/single-spa/single-spa-dojo/issues/5 and https://github.com/dojo/framework/issues/676 and 
image (1)

The memory leak / partial unmount isn't great, but is captured in https://github.com/single-spa/single-spa-dojo/issues/5 which I think is okay for now while we wait on dojo to add this.

filoxo (approved):

filoxo:
Awesome, good to know! I hadn't watched this very closely. Just surprising how quickly it came together!

PR #190, opened 2020-03-02
Updating svelte docs for single-spa-svelte@2.0.0.

internettrans:
See related https://github.com/single-spa/single-spa-svelte/pull/6


TheMcMurder (approved):

filoxo (commented):
I believe the difference between `props` and `data` is just that the former is for Svelte 3 and the latter for Svelte 2 and below. Not sure if that's worth mentioning.

PR #191, opened 2020-03-03
Fixing build failures

internettrans:
This solves root cause of https://github.com/single-spa/single-spa.js.org/issues/164 (I think)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,10 +33,11 @@ workflows:
     jobs:
       - deploy-website:
           filters:
-            ignore:
-              - gh-pages
             branches:
               only: master
   build:
     jobs:
-      - build-website
+      - build-website:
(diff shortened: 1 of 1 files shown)


internettrans (commented):
https://circleci.com/docs/2.0/configuration-reference/#branches

frehner:
hm, I'm confused - the `only` should just work there and already ignore the `gh-pages` branch shouldn't it?

internettrans:
Ah good point. I should have put this on the `build-website` job, not the `deploy-website` job. [Here is an example failure](https://circleci.com/gh/single-spa/zh-hans.single-spa.js.org/42) -- I think it's on the build-website job since there is no package json in the gh-pages branch.

I have updated this to be on the correct job

frehner (approved):

PR #200, opened 2020-03-10
fix: api.md html snippet render error fix.

billfeller:
1. use &lt;codehtml/> instead of &lt;code/>  in html snippet.
2. use {/*  */} instead of &lt;!-- --> in html snippet.

### DIFF
--- a/website/src/theme/MDXComponents/index.js
+++ b/website/src/theme/MDXComponents/index.js
@@ -1,0 +1,40 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React from 'react';
+import Link from '@docusaurus/Link';
+import CodeBlock from '@theme/CodeBlock';
+import Heading from '@theme/Heading';
+import styles from './styles.module.css';
+
+export default {
+  codehtml: props => {
+    const {children} = props;
+    return <code>{children}</code>
+  },
+  code: props => {
+    const {children} = props;
+    if (typeof children === 'string') {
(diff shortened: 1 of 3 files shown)


frehner:
I’m good with the change to comments

But would like to keep the code different from normal text. Since this is markdown, we should be able to use backticks to replace it, right?

billfeller:
yeah, i have tried backticks yet, but it can not be rendered in html snippet as below.

image

i try &lt;b> as below, it look good enough to emphasize the words.

image

which one to choose, backticks or &lt;b>

internettrans:
Might be best to switch off of html entirely and just use markdown

frehner:
An alternative would be to add a class to `code` blocks that allows them to either be as they are or optionally an inline block of text (like what is desired here). I think @filoxo is the one that originally did the work on this - thoughts?

internettrans:
I actually think it was @alocke12992 who might have put this documentation together. Not entirely sure

frehner:
> I actually think it was @alocke12992 who might have put this documentation together. Not entirely sure

I meant @filoxo was the person that originally coded/styled the `code` tags I believe.

billfeller:
i try to extend `src/theme/MDXComponents` to use &lt;code/>. so it need to use &lt;codehtml/>in html snippet instead, then we can use originally coded/styled.

image

i just use swizzle the [theme-components](https://v2.docusaurus.io/docs/using-themes/#swizzling-theme-components) and extend with codehtml.  68549fba1a60e75794139c883d1e035e4a7242af

@frehner @joeldenning

billfeller:
```suggestion
  codehtml: props => {
    const {children} = props;
    return <code>{children}</code>
  },
```

frehner (approved):
That looks great, thanks for your work on this!

PR #203, opened 2020-03-17
Adding documentation about Module Federation.

internettrans:
image


TheMcMurder (approved):
This looks good. Hopefully this will help people choose an approach. I’m surprised at how so many people are excited about module federation instead of the import map approach. As far as I’m aware there isn’t a compelling reason to choose module federation over import maps 🤷🏼‍♂️

TheMcMurder:
> image
> 
> 

Thanks for including the screen shot. It made it much easier to review.

frehner:
The advantage I see is that this still provides a way to have “standalone” apps. 

How often people will use that in practice I’m not sure though

internettrans:
I think people like standalone apps for development, but not for production. Might be worthwhile adding it to the docs, though

PR #207, opened 2020-03-23
Clarification on parcels

TheMcMurder:
This new page should help us clarify the difference between parcels, applications, and utilities.

### DIFF
--- a/website/versioned_docs/version-5.x/microfrontends-concept.md
+++ b/website/versioned_docs/version-5.x/microfrontends-concept.md
@@ -36,7 +36,7 @@ In the context of single-spa, there are three kinds of microfrontends:
 2. [single-spa parcels](/docs/parcels-overview): Microfrontends that render components without controlling routes.
 3. [utility modules](/docs/recommended-setup#utility-modules-styleguide-api-etc): Microfrontends that export shared javascript logic, without rendering components.
 
-A web app may include one or more types of microfrontends. [Choosing between microfrontend types](/docs/recommended-setup#applications-versus-parcels-versus-utility-modules).
+A web app may include one or more types of microfrontends. See the [differences broken down](/docs/module-types), and our recommendations on [choosing between microfrontend types](/docs/recommended-setup#applications-versus-parcels-versus-utility-modules).
 
 ## Communication between Microfrontends
 
--- a/website/versioned_docs/version-5.x/microfrontends-concept.md
+++ b/website/versioned_docs/version-5.x/microfrontends-concept.md
@@ -36,7 +36,7 @@ In the context of single-spa, there are three kinds of microfrontends:
 2. [single-spa parcels](/docs/parcels-overview): Microfrontends that render components without controlling routes.
 3. [utility modules](/docs/recommended-setup#utility-modules-styleguide-api-etc): Microfrontends that export shared javascript logic, without rendering components.
 
-A web app may include one or more types of microfrontends. See the [differences broken down](/docs/module-types), and our recommendations on [choosing between microfrontend types](/docs/recommended-setup#applications-versus-parcels-versus-utility-modules).
+A web app may include one or more types of microfrontends. See [an in-depth comparison](/docs/module-types), and our recommendations for [choosing between microfrontend types](/docs/recommended-setup#applications-versus-parcels-versus-utility-modules).
 
 ## Communication between Microfrontends
 
--- a/website/versioned_docs/version-5.x/module-types.md
+++ b/website/versioned_docs/version-5.x/module-types.md
@@ -12,7 +12,7 @@ Here is how each single-spa microfrontend works conceptually. This information s
 
 | application                       | parcel                               | utility                              |
 | --------------------------------- | ------------------------------------ | ------------------------------------ |
-| has multiple routes               | no routes                            | no route                             |
+| has multiple routes               | no routes                            | no routes                            |
 | declarative API                   | imperative API                        | not registered/mounted by single-spa |
 | renders UI                        | renders UI                           | no UI                                |
 | single-spa managed lifecylces     | custom managed lifecycles            | no DOM elements                      |
--- a/website/versioned_docs/version-5.x/module-types.md
+++ b/website/versioned_docs/version-5.x/module-types.md
@@ -1,7 +1,7 @@
 ---
 id: module-types
 title: single-spa microfrontend types
-sidebar_label: Conceptual Module Types
+sidebar_label: Microfrontend Types
 ---
 
 # Concept: single-spa microfrontend types
--- a/website/versioned_docs/version-5.x/module-types.md
+++ b/website/versioned_docs/version-5.x/module-types.md
@@ -19,3 +19,33 @@ Here is how each single-spa module works conceptually. This information should h
 | Core building block               | only needed with multiple frameworks | useful to share business logic       |
 
 Each single-spa module is in reality an in browser javascript module see [here](/docs/recommended-setup#in-browser-versus-build-time-modules) for more information. Each type/module fits into the overal microfrontend approach.
+
+## Applications
+
+### Applications are declarative
+Applications use a declarative api. Your single-spa config (also sometimes called root config) defines applications ahead of time, defines the condition at which they are active, but doesn't mount the application directly.
+
+### Applications have managed lifecylces
+single-spa manages registered applications and is in charge of all of their lifecycles. This prevents you from writing a bunch of logic about when applications should mount and unmount; single-spa takes care of that for you.
+All single spa needs to make this work automatically is for an activity function that describes when your application should be active.
+
+## Parcels
+
+### Parcels are impertive
+Parcels exist in many ways as an escape hatch from the normal declarative flow. They exist primarily to allow you to reuse UI across applications when those applications are written in multiple frameworks.
+
+### You manage the lifecycles of Parcels
+When you call `mountParcel` or `mountRootParcel` the parcel is mounted immediately and returns the parcel object. You need to call the `unmount` method on the parcel manually when the component that calls `mountParcel` unmounts.
+
+### Parcels are best suited for sharing components between frameworks
+Taking a component and turning it into a `parcelConfig` is as easy as using the [single-spa helpers](/docs/ecosystem#help-for-frameworks) for that framework. This results in an object that single-spa can use to create and mount a parcel.
+Because single-spa can mount a parcel anywhere, this gives you a way to share components across frameworks. It should not be used if the shared component is being used in another application of the same framework.
+For example: `applicationOne` is written in Vue and contains all the UI/Logic to create a user. `application2` is written in react and needs to create a user. Using a single-spa parcel allows you to wrap your `app1` vue component
+in a way that will make it work inside `application2` despite the different frameworks. Even better if `application2` is unmounted by single-spa (per the activity function returning false)
+Think of parcels as a single-spa specific implementation of webcomponents.
+
+## Utility modules share business logic
+Utility modules are a great place to share business logic. Instead of each application creating their own implementation of something tied to your buisness logic, you can use a plain javascript object (single-spa utility) to share that logic.
+For example: Authorization. How does each application know which user is logged in? You could have each application ask the server or read a JWT but that creates duplicate work in each application.
+Instead implmenting logic around who is logged in one module (with all the necessary methods exported on the module) and each application can simply ask the utility module who is logged in.
+This approach also works well for data [fetching](/docs/recommended-setup#api-data).
--- a/website/versioned_docs/version-5.x/module-types.md
+++ b/website/versioned_docs/version-5.x/module-types.md
@@ -1,0 +1,21 @@
+---
+id: module-types
+title: single-spa conceptual module types
+sidebar_label: Conceptual Module Types
+---
+
+# Concept: single-spa modules
+
+Single-spa has [different categories](/docs/microfrontends-concept#types-of-microfrontends) of microfrontends. It is up to you where and how you use each of them. However, the single-spa core team has [recommendations](/docs/recommended-setup/#applications-versus-parcels-versus-utility-modules).
+
+Here is how each single-spa module works conceptually. This information should help you understand our [recommendations](/docs/recommended-setup/#applications-versus-parcels-versus-utility-modules) on which to use where.
+
+| application                       | parcel                               | utility                              |
+| --------------------------------- | ------------------------------------ | ------------------------------------ |
+| has multiple routes               | no routes                            | no route                             |
+| declarative API                   | impertive API                        | not registered/mounted by single-spa |
+| renders UI                        | renders UI                           | no UI                                |
+| single-spa managed lifecylces     | custom managed lifecycles            | no dom elements                      |
+| Core building block               | only needed with multiple frameworks | useful to share business logic       |
+
+Each single-spa module is in reality an in browser javascript module see [here](/docs/recommended-setup#in-browser-versus-build-time-modules) for more information. Each type/module fits into the overal microfrontend approach.
--- a/website/versioned_docs/version-5.x/module-types.md
+++ b/website/versioned_docs/version-5.x/module-types.md
@@ -6,19 +6,19 @@ sidebar_label: Microfrontend Types
 
 # Concept: single-spa microfrontend types
 
-Single-spa has [different categories](/docs/microfrontends-concept#types-of-microfrontends) of microfrontends. It is up to you where and how you use each of them. However, the single-spa core team has [recommendations](/docs/recommended-setup/#applications-versus-parcels-versus-utility-modules).
+Single-spa has [different categories](./microfrontends-concept#types-of-microfrontends) of microfrontends. It is up to you where and how you use each of them. However, the single-spa core team has [recommendations](./recommended-setup/#applications-versus-parcels-versus-utility-modules).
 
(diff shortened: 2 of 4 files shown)


website/versioned_docs/version-5.x/module-types.md
```diff
@@ -0,0 +1,51 @@
+---
+id: module-types
+title: single-spa conceptual module types
+sidebar_label: Conceptual Module Types
+---
+
+# Concept: single-spa modules
```
TheMcMurder: I'm torn on using the term `module` I worry that it's overloaded but I didn't like `types` either for the same reason, if anyone has a suggestion on a better verb please let me know.

website/versioned_docs/version-5.x/parcels-overview.md
```diff
@@ -3,7 +3,8 @@ id: parcels-overview
 title: Parcels
 sidebar_label: Overview
 ---
-_Parcels are an advanced feature of single-spa. Avoid using them until you have a better understanding of single-spa's registerApplication api._
+_Parcels are an advanced feature of single-spa. Avoid using them until you have a better understanding of [single-spa module types](/docs/module-types)._
+
```
TheMcMurder, line 7: Changing to point to the differences made sense to me. I also added a line of whitespace to help differentiate this line.

website/versioned_docs/version-5.x/module-types.md
filoxo: I agree, module doesn't seem right for our docs. I think it's be best to stay consistent with our current "single-spa applications", or lean into the "microfrontend" or "browser microservice" terminology.

website/versioned_docs/version-5.x/module-types.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Parcels
+
+### Parcels are imperitive
+Parcels exist in many ways as an escape hatch from the normal declarative flow. They exist primarily to allow you to reuse UI across applications when those applications are written in multiple frameworks.
+
+### You manage the lifecycles of Parcels
+When you call `mountParcel` or `mountRootParcel` [(see api)](/docs/parcels-api) the parcel is mounted immediately and returns the parcel object. You need to call the `unmount` method on the parcel manually when the component that calls `mountParcel` unmounts.
+
+### Parcels are best suited for sharing components between frameworks
```
filoxo: "components" seems a bit too overloaded too. What do you think of

> best suited for sharing UI between frameworks.

Alternative terms I think should be considered are "widgets" or "UI components".

website/versioned_docs/version-5.x/parcels-overview.md
filoxo: This seems overly... strict? What about something along the lines of "Parcels are an advanced feature that is almost never needed until X, Y, Z. Therefore, you should not need to reach for parcels until your architecture has matured"? Just a thought.

TheMcMurder: I see far too many people using parcels instead of applications. I want to push people to use applications as much as possible. In the long run using parcels for everything is going to cause problems. 

I’ve seen several POC implementations of single-spa that use parcels _instead_ of applications. I want to make it very clear that parcels are not where development effort should be focused unless you actually know you need them. 

A lot of learners reach for parcels and ignore the application API entirely. 

I don’t know how to tell people to go somewhere else and come back to parcels later.

website/versioned_sidebars/version-5.x-sidebars.json
```diff
@@ -88,6 +88,16 @@
         }
       ]
     },
+    {
+      "type": "category",
+      "label": "Concept: Single spa modules",
```
internettrans: Honestly, I think this should go under the "Concept: Microfrontends" label, instead of a whole new one. To me this is just an extension of the microfrontend concept. We could rename `overview.md` inside of there to something else if it helps, but I think it would be good to have more links under our existing labels instead of just one link per label

website/versioned_docs/version-5.x/microfrontends-concept.md
```diff
@@ -36,7 +36,7 @@ In the context of single-spa, there are three kinds of microfrontends:
 2. [single-spa parcels](/docs/parcels-overview): Microfrontends that render components without controlling routes.
 3. [utility modules](/docs/recommended-setup#utility-modules-styleguide-api-etc): Microfrontends that export shared javascript logic, without rendering components.
 
-A web app may include one or more types of microfrontends. [Choosing between microfrontend types](/docs/recommended-setup#applications-versus-parcels-versus-utility-modules).
+A web app may include one or more types of microfrontends. See the [differences broken down](/docs/module-types), and our recommendations on [choosing between microfrontend types](/docs/recommended-setup#applications-versus-parcels-versus-utility-modules).
```
internettrans: nit: "See the differences broken down" is a bit harder for non-native english speakers and for our translators:

```suggestion
A web app may include one or more types of microfrontends. See [an in-depth comparison](/docs/module-types), and our recommendations for [choosing between microfrontend types](/docs/recommended-setup#applications-versus-parcels-versus-utility-modules).
```

website/versioned_docs/version-5.x/module-types.md
internettrans: Yeah we've already introduced these concepts in other places with other terminology, where we call it "types of microfrontends". The term "single-spa modules" isn't my favorite.

Honestly I wonder if this whole new page would be better off just as additional sections to microfrontends-concept.md? It's only 50 lines of documentation, which isn't a ton.

website/versioned_docs/version-5.x/module-types.md
```diff
@@ -0,0 +1,51 @@
...
+# Concept: single-spa modules
+
+Single-spa has [different categories](/docs/microfrontends-concept#types-of-microfrontends) of microfrontends. It is up to you where and how you use each of them. However, the single-spa core team has [recommendations](/docs/recommended-setup/#applications-versus-parcels-versus-utility-modules).
+
+Here is how each single-spa module works conceptually. This information should help you understand our [recommendations](/docs/recommended-setup/#applications-versus-parcels-versus-utility-modules).
+
+| application                       | parcel                               | utility                              |
+| --------------------------------- | ------------------------------------ | ------------------------------------ |
+| has multiple routes               | no routes                            | no route                             |
+| declarative API                   | impertive API                        | not registered/mounted by single-spa |
```
internettrans:
```suggestion
| declarative API                   | imperative API                        | not registered/mounted by single-spa |
```

website/versioned_docs/version-5.x/module-types.md
```diff
@@ -0,0 +1,51 @@
...
+
+# Concept: single-spa modules
+
+Single-spa has [different categories](/docs/microfrontends-concept#types-of-microfrontends) of microfrontends. It is up to you where and how you use each of them. However, the single-spa core team has [recommendations](/docs/recommended-setup/#applications-versus-parcels-versus-utility-modules).
+
+Here is how each single-spa module works conceptually. This information should help you understand our [recommendations](/docs/recommended-setup/#applications-versus-parcels-versus-utility-modules).
+
+| application                       | parcel                               | utility                              |
+| --------------------------------- | ------------------------------------ | ------------------------------------ |
+| has multiple routes               | no routes                            | no route                             |
```
internettrans:
```suggestion
| has multiple routes               | no routes                            | no routes                            |
```

website/versioned_docs/version-5.x/module-types.md
```diff
@@ -0,0 +1,51 @@
...
+| declarative API                   | impertive API                        | not registered/mounted by single-spa |
+| renders UI                        | renders UI                           | no UI                                |
+| single-spa managed lifecylces     | custom managed lifecycles            | no DOM elements                      |
```
internettrans:
```suggestion
| single-spa managed lifecycles     | custom managed lifecycles            | no DOM elements                      |
```

One thing - it is not strictly true that utility modules do not deal with dom elements. A styleguide, for example, exposes UI components that create DOM elements. It sounds like we might be talking about lifecycles or "mount points" in this table row, so the languaging for utility modules could be "no lifecycle methods"

website/versioned_docs/version-5.x/module-types.md
```diff
@@ -0,0 +1,51 @@
...
+| Core building block               | only needed with multiple frameworks | useful to share business logic       |
```
internettrans: Honestly I don't like the term business logic much, because I find it to be a buzz term / catch-all that means whatever anyone wants it to mean. Also, I think utility modules can do a lot more than just business logic.

website/versioned_docs/version-5.x/module-types.md
```diff
@@ -0,0 +1,51 @@
...
+
+Each single-spa module is in reality an in browser javascript module see [here](/docs/recommended-setup#in-browser-versus-build-time-modules) for more information. Each type/module fits into the overall microfrontend approach.
```
internettrans:
```suggestion
Each microfrontend is an in-browser javascript module ([explanation](/docs/recommended-setup#in-browser-versus-build-time-modules)).
```

website/versioned_docs/version-5.x/module-types.md
```diff
@@ -0,0 +1,51 @@
...
+### Applications are declarative
+Applications use a declarative api. Your single-spa config (also sometimes called root config) defines applications ahead of time, defines the condition at which they are active, but doesn't mount the application directly.
+
+### Applications have managed lifecycles
+single-spa manages registered applications and is in charge of all of their lifecycles. This prevents you from writing a bunch of logic about when applications should mount and unmount; single-spa takes care of that for you.
+All single spa needs to make this work automatically is for an activity function that describes when your application should be active.
+
+## Parcels
+
+### Parcels are imperitive
```
internettrans:
```suggestion
### Parcels are imperative
```

website/versioned_docs/version-5.x/module-types.md
internettrans: I think that the term component works fine here. Parcels really are framework agnostic components.

website/versioned_docs/version-5.x/module-types.md
```diff
@@ -0,0 +1,51 @@
...
+For example: `applicationOne` is written in Vue and contains all the UI/Logic to create a user. `application2` is written in React and needs to create a user. Using a single-spa parcel allows you to wrap your `app1` Vue component
+in a way that will make it work inside `application2` despite the different frameworks. Even better if `application2` is unmounted by single-spa (per the activity function returning false)
+Think of parcels as a single-spa specific implementation of webcomponents.
+
+## Utility modules share business logic
+Utility modules are a great place to share business logic. Instead of each application creating their own implementation of something tied to your buisness logic, you can use a plain javascript object (single-spa utility) to share that logic.
+For example: Authorization. How does each application know which user is logged in? You could have each application ask the server or read a JWT but that creates duplicate work in each application.
+Instead implmenting logic around who is logged in one module (with all the necessary methods exported on the module) and each application can simply ask the utility module who is logged in.
```
internettrans:
```suggestion
Instead of implementing logic around who is logged in one module (with all the necessary methods exported on the module) and each application can simply ask the utility module who is logged in.
```

website/versioned_docs/version-5.x/parcels-overview.md
internettrans: I found the languaging here to be a bit aggressive / condescending. "Wait until you understand better" is assuming something about the user that might not be true.

```suggestion
_Parcels are an advanced feature of single-spa. We recommend that you use applications as the primary type of microfrontend in your architecture. See [this explanation](/docs/module-types) for more details
```

website/versioned_docs/version-5.x/module-types.md
```diff
@@ -0,0 +1,51 @@
...
+sidebar_label: Conceptual Module Types
+---
+
+# Concept: single-spa microfrontend types
+
+Single-spa has [different categories](/docs/microfrontends-concept#types-of-microfrontends) of microfrontends. It is up to you where and how you use each of them. However, the single-spa core team has [recommendations](/docs/recommended-setup/#applications-versus-parcels-versus-utility-modules).
+
+Here is how each single-spa microfrontend works conceptually. This information should help you understand our [recommendations](/docs/recommended-setup/#applications-versus-parcels-versus-utility-modules).
+
+| application                       | parcel                               | utility                              |
```
internettrans: I think this table would make more sense if there were a column that explained what each row was. For example, the first row could be called "routing".

```suggestion
| Topic                | application                       | parcel                               | utility                              |
```

TheMcMurder:
Screen Shot 2020-03-24 at 9 43 12 AM

website/versioned_docs/version-5.x/module-types.md
```diff
@@ -0,0 +1,51 @@
+---
+id: module-types
+title: single-spa microfrontend types
+sidebar_label: Microfrontend Types
+---
+
+# Concept: single-spa microfrontend types
+
+Single-spa has [different categories](/docs/microfrontends-concept#types-of-microfrontends) of microfrontends. It is up to you where and how you use each of them. However, the single-spa core team has [recommendations](/docs/recommended-setup/#applications-versus-parcels-versus-utility-modules).
```
internettrans: Oh one thing I just remembered - we should start using relative links for linking between files. This is important for versioning - we want to always stay within the same version of the single-spa docs when linking. See related https://github.com/single-spa/single-spa.js.org/issues/178 and https://v2.docusaurus.io/docs/versioning/#linking-files

```suggestion
Single-spa has [different categories](/docs/microfrontends-concept#types-of-microfrontends) of microfrontends. It is up to you where and how you use each of them. However, the single-spa core team has [recommendations](./recommended-setup.md#applications-versus-parcels-versus-utility-modules).
```

Please modify all links in this page accordingly

website/versioned_docs/version-5.x/module-types.md
```diff
@@ -0,0 +1,51 @@
...
+
+Here is how each single-spa microfrontend works conceptually. This information should help you understand our [recommendations](/docs/recommended-setup/#applications-versus-parcels-versus-utility-modules).
+
+| Topic                | application                       | parcel                               | utility                              |
+| -------------------- | --------------------------------- | ------------------------------------ | ------------------------------------ |
+| routing              | has multiple routes               | no routes                            | no routes                            |
+| API                  | declarative API                   | imperative API                       | no single-spa api                    |
+| renders UI           | renders UI                        | renders UI                           | doesn't directly render UI           |
+| lifecycles           | single-spa managed lifecylces     | custom managed lifecycles            | no lifecycles                        |
```
internettrans:
```suggestion
| lifecycles           | single-spa managed lifecycles     | custom managed lifecycles            | no lifecycles                        |
```

website/versioned_docs/version-5.x/module-types.md
```diff
@@ -0,0 +1,51 @@
...
+| When to use          | Core building block               | only needed with multiple frameworks | useful to share common logic         |
+
+Each single-spa microfrontend is an in-browser javascript module ([explaination](/docs/recommended-setup#in-browser-versus-build-time-modules)).
+
+## Applications
+
+### Applications are declarative
+Applications use a declarative api. Your single-spa config (also sometimes called root config) defines applications ahead of time, defines the condition at which they are active, but doesn't mount the application directly.
```
internettrans: Maybe clarify that `registerApplication` is the declarative API?

```suggestion
Applications use a declarative api called `registerApplication`. Your single-spa config (also sometimes called root config) defines applications ahead of time, defines the condition at which they are active, but doesn't mount the application directly.
```

website/versioned_docs/version-5.x/module-types.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Parcels
+
+### Parcels are imperative
+Parcels exist in many ways as an escape hatch from the normal declarative flow. They exist primarily to allow you to reuse UI across applications when those applications are written in multiple frameworks.
+
+### You manage the lifecycles of Parcels
+When you call `mountParcel` or `mountRootParcel` [(see api)](/docs/parcels-api) the parcel is mounted immediately and returns the parcel object. You need to call the `unmount` method on the parcel manually when the component that calls `mountParcel` unmounts.
+
+### Parcels are best suited for sharing UI between frameworks
```
internettrans, line 40: I think some people use parcels because they don't know what the alternative is. Should we perhaps have another subsection called "Alternatives to parcels" - maybe linking to the react styleguide that exports react components directly into other applications.

website/versioned_docs/version-5.x/module-types.md
```diff
@@ -0,0 +1,51 @@
+---
+id: module-types
+title: single-spa microfrontend types
+sidebar_label: Microfrontend Types
+---
+
+# Concept: single-spa microfrontend types
+
+Single-spa has [different categories](./microfrontends-concept#types-of-microfrontends) of microfrontends. It is up to you where and how you use each of them. However, the single-spa core team has [recommendations](./recommended-setup/#applications-versus-parcels-versus-utility-modules).
```
internettrans: [The docusaurus docs](https://v2.docusaurus.io/docs/versioning/#linking-files) suggest putting in `.md` for relative links, but the extension is not here. Have you verified whether omitting the `.md` still works?

TheMcMurder: Yeah it does work. I added the `.md` anyways though

website/versioned_docs/version-5.x/module-types.md
TheMcMurder: That's a good idea. I don't have an example of that yet (but will soon) I created [an issue](https://github.com/single-spa/single-spa.js.org/issues/208) for it so I don't forget and put that in another MR

internettrans (approved):
Nice work

PR #209, opened 2020-03-24
Adding prominent slack chat link.

internettrans:
I think the slack workspace is important enough to a lot of people's success that it should be advertised more prominently on the home page. Do you agree?

image


TheMcMurder (approved):
🐙

PR #210, opened 2020-03-25
fix mistake with customProps

aknovoroll:
you should use props.authToken instead of props.customProps.authToken


frehner (approved):
ah yeah. if I remember correctly, we originally put dev-supplied props on customProps, but then just decided to merge them into props. 

Thanks.

PR #216, opened 2020-04-02
Add object config docs

Sauloxd:
- Add documentation for configuration object on `registerApplication`
- Add error 38, regarding the accept keys in the configuration object (pending PR to send valid keys and invalid keys to improve error UX)
- Add error 39, saying it only accepts a configuration object which is not an array and not null

### DIFF
--- a/website/src/components/Errors/Codes/38.js
+++ b/website/src/components/Errors/Codes/38.js
@@ -1,7 +1,6 @@
-import React from 'react'
+import React from 'react';
 
-export default function ErrorCode37(props) {
-
+export default function ErrorCode38(props) {
   return (
     <>
       <h1>#38: Invalid registerApplication configuration object</h1>
@@ -15,5 +14,5 @@ export default function ErrorCode37(props) {
         See <a href="/docs/configuration">registerApplication API documentation</a> for more information.
       </p>
     </>
-  )
+  );
 }
--- a/website/src/components/Errors/Codes/38.js
+++ b/website/src/components/Errors/Codes/38.js
@@ -1,0 +1,19 @@
+import React from 'react'
+
+export default function ErrorCode37(props) {
+
+  return (
+    <>
+      <h1>#38: Invalid registerApplication configuration object</h1>
+      <p>
+        The configuration object in registerApplication accepts only the following keys: "{props.getErrorCodeArg(0)}"
+      </p>
+      <p>
+        You provided the following invalid keys: "{props.getErrorCodeArg(1)}"
+      </p>
+      <p>
+        See <a href="/docs/configuration">registerApplication API documentation</a> for more information.
+      </p>
+    </>
+  )
+}
--- a/website/src/components/Errors/Codes/39.js
+++ b/website/src/components/Errors/Codes/39.js
@@ -1,0 +1,16 @@
+import React from 'react'
+
+export default function ErrorCode37() {
+
+  return (
+    <>
+      <h1>#39: Invalid registerApplication configuration object</h1>
+      <p>
+        Configuration object can't be an array or null! Please provide a plain object with correct configuration.
+      </p>
+      <p>
+        See <a href="/docs/configuration">registerApplication API documentation</a> for more information.
+      </p>
+    </>
+  )
+}
--- a/website/src/components/Errors/Codes/39.js
+++ b/website/src/components/Errors/Codes/39.js
@@ -1,7 +1,6 @@
-import React from 'react'
+import React from 'react';
 
-export default function ErrorCode37() {
-
+export default function ErrorCode39() {
   return (
     <>
       <h1>#39: Invalid registerApplication configuration object</h1>
@@ -12,5 +11,5 @@ export default function ErrorCode37() {
         See <a href="/docs/configuration">registerApplication API documentation</a> for more information.
       </p>
     </>
-  )
+  );
 }
--- a/website/versioned_docs/version-5.x/api.md
+++ b/website/versioned_docs/version-5.x/api.md
@@ -4,7 +4,7 @@ title: Applications API
 sidebar_label: Applications API
 ---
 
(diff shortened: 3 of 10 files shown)


website/src/components/Errors/Codes/38.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react'
+
+export default function ErrorCode37(props) {
```
internettrans:
```suggestion
export default function ErrorCode38(props) {
```

website/src/components/Errors/Codes/39.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react'
+
+export default function ErrorCode37() {
```
internettrans:
```suggestion
export default function ErrorCode39() {
```

website/versioned_docs/version-5.x/api.md
```diff
@@ -42,6 +44,39 @@ Note that if an application is registered from within another application, that
...
+	app: () => System.import('appName'),
+	activeWhen: '/appName'
+	customProps: {}
+})
+```
+
+<h3>arguments</h3>
+
+<dl className="args-list">
+	<dt>name?: string</dt>
```
internettrans: The question mark here makes it seem like name is optional, but it isn't. Please remove:

```suggestion
	<dt>name: string</dt>
```

Same goes for the other properties that are required. The `customProps` are optional, but everything else is required.

website/versioned_docs/version-5.x/api.md
```diff
@@ -42,6 +44,39 @@ Note that if an application is registered from within another application, that
...
+	<dd>App name that single-spa will register and reference this application with, and will be labelled with in dev tools.</dd>
+	<dt>app?: Application | () => Application | Promise&lt;Application&gt; </dt>
+	<dd>Application object or a function that returns the resolved application (Promise or not)</dd>
+	<dt>activeWhen: string | (location) => boolean | (string | (location) => boolean)[]</dt>
+	<dd>Can be a path prefix which will match every URL starting with this path,
```
internettrans, line 65: Let's add to this some explanation about dynamic path params in the string. I see you have some detail for it on the other page, but this page should at least mention it.

website/versioned_docs/version-5.x/single-spa-config.md
```diff
@@ -82,6 +89,38 @@ single-spa will call each application's activity function under the following sc
...
+
+#### config.name
+Must be a string name
+
+#### config.app
+The definition of your app, which can be an object with single-spa lifecycle
+methods, or a loading function, the same as the second argument on the arguments API
+
+#### config.activeWhen
+Can be an activity function, like the arguments API, a path prefix or and array
```
internettrans:
```suggestion
Can be an activity function, like the arguments API, a path prefix or an array
```

internettrans (commented):
Thanks for your work on this - just a few small comments.

Also, I wouldn't mind if we updated other places in the documentation to use the object version of the API, as the default. Also, we could switch over the main example repos to use it, too.

website/versioned_docs/version-5.x/building-applications.md
```diff
@@ -61,12 +61,12 @@ In addition to the built-in props that are provided by single-spa, you may optio
...
-  'app1', 
-  () => {}, 
-  () => {}, 
-  { authToken: "d83jD63UdZ6RS6f70D0" }
-);
+singleSpa.registerApplication({
+  name: 'app1',
+  activeWhen,
+  app,
+  customProps: { authToken: "d83jD63UdZ6RS6f70D0" }
```
filoxo, line 68: I hope this isn't too nitpicky, but could the order of these properties match the API documentation (app before activeWhen)? Consistency is very helpful in documentation. 

Also, shouldn't `activeWhen` and `app` be defined with something? Neither are defined as variables so this would throw an error if executed.

website/versioned_docs/version-5.x/single-spa-config.md
```diff
@@ -25,31 +26,37 @@ does not have to. Note that if an application is registered from within another
...
 
-function loadingFunction() {
-  return import("src/app1/main.js");
-}
+// Config with more expressive API
+registerApplication({
+    name: 'app1',
+    app: () => import('src/app1/main.js'),
+    activeWhen: '/app1'
+);
```
filoxo, line 47: Missing closing brace `}`

internettrans (approved):
Thanks for your help on this - looks good to me

filoxo:
haha looks like I submitted my review seconds after Joel merged it.

internettrans:
Oops yeah, @Sauloxd could you do a follow-up PR to address these comments from @filoxo ?

website/versioned_docs/version-5.x/api.md
```diff
@@ -4,7 +4,7 @@ title: Applications API
 sidebar_label: Applications API
 ---
 
-single-spa exports named functions and variables rather than a single default export.
+Single-spa exports named functions and variables rather than a single default export.
```
TheMcMurder, line 7: I was under the impression we wanted to keep: `single-spa` lowercase even at the beginning of a sentence, is that no longer true?

filoxo: That's what I've been doing but we should probably create a STYLEGUIDE.md file to put those rules down in writing.

Sauloxd: Sorry, I thought that was not intended! Creating a doc would help prevent contributors to do silly mistakes like this 👍

website/versioned_docs/version-5.x/building-applications.md
Sauloxd: not at all! I'd do this changes too in a later PR. The more professional the document looks, the better!
I'll take this in consideration in the next PR fixes :)

PR #218, opened 2020-04-02
Adding link to gist that shows how to use react-app-rewired.

internettrans:
Related - https://github.com/single-spa/single-spa/issues/503


filoxo (commented):
Would it be good to also note that this is "unofficial" or only "cursory" support? Given how many questions we get regarding CRA, I worry that this will be seen as "official support".

internettrans:
👍 added some languaging. Let me know what you think of it.

filoxo (approved):
👍

PR #227, opened 2020-04-13
Updating docs for create-single-spa@1.5

internettrans:
See related https://github.com/single-spa/create-single-spa/releases/tag/v1.5.0


website/versioned_docs/version-5.x/create-single-spa.md
```diff
@@ -131,10 +131,6 @@ module.exports = webpackConfigEnv => {
     orgName: 'name-of-company',
     // The name of the current project. This usually matches the git repo's name
     projectName: 'name-of-project',
-    // If your project uses typescript, set typecheck. Otherwise, omit it.
-    typecheck: 'typescript',
```
internettrans, line 135: This option is no longer supported in webpack-config-single-spa. Instead, webpack-config-single-spa-ts should be used instead. This is technically a breaking change that I snuck into create-single-spa@1.5.0 since I didn't think anyone would notice because it had only been released for a few days.

website/versioned_docs/version-5.x/create-single-spa.md
```diff
@@ -131,10 +131,6 @@ module.exports = webpackConfigEnv => {
...
-    // Set to true if your project uses react
-    react: true,
```
internettrans, line 137: This option is no longer supported in webpack-config-single-spa. Instead, webpack-config-single-spa-react-ts should be used instead. This is technically a breaking change that I snuck into create-single-spa@1.5.0 since I didn't think anyone would notice because it had only been released for a few days.

PR #232, opened 2020-04-19
Fix typo

sniperwolf:
Removed repeated for

### DIFF
--- a/website/versioned_docs/version-5.x/ecosystem-angularjs.md
+++ b/website/versioned_docs/version-5.x/ecosystem-angularjs.md
@@ -4,7 +4,7 @@ title: single-spa-angularjs
 sidebar_label: AngularJS
 ---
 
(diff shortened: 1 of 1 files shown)


frehner (approved):
Thanks!

verinder-bindra:
Hi sniperwolf, I am facing issues using directives  from angularjs app running under single spa. when directive templateUrl is used... LoadTemplate is failing because it tries to find the template under root config url. There is a solution for Angular Apps, what about AngularJS?  How assets can be exported... i am using parcel instead of webpack. Note when i use Template property of directive, it all works . I tried running angularjs app indipendently.. it works too with templateUrl, but not under single spa root config.  Please add documentation for that as well.
please look at stackoverflow article explaining the issue i have created :   https://stackoverflow.com/questions/73313634/single-spa-angularjs-microfrontend-not-loading-html-templateurl-for-directives

PR #247, opened 2020-05-01
Adding documentation for before-mount-routing-event.

internettrans:
See related https://github.com/single-spa/single-spa/pull/534

### DIFF
--- a/website/versioned_docs/version-5.x/api.md
+++ b/website/versioned_docs/version-5.x/api.md
@@ -629,7 +629,7 @@ window.addEventListener('single-spa:before-mount-routing-event', (evt) => {
 });
 ```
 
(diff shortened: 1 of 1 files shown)


website/versioned_docs/version-5.x/api.md
```diff
@@ -620,6 +620,18 @@ window.addEventListener('single-spa:before-routing-event', () => {
...
+## before mount routing event
+
+```js
+window.addEventListener('single-spa:before-mount-routing-event', (evt) => {
+	console.log('single-spa is about to mount/unmount applications!');
+	console.log(evt.detail)
+});
+```
+
+A `single-spa:before-mount-routing-event` event is fired after `before-routing-event` and before `routing-event`. It is guaranteed to fire after all single-spa applications have been unmounted, but before any new applications have been mounted.
```
frehner, line 632: small nit but I think this would read slightly easier if you put the `before` one first and the `after` one second.

internettrans: 👍 updated

frehner (approved):
sorry about that

PR #252, opened 2020-05-07
Fixing docs for single-spa routing evt properties

internettrans:
See related https://github.com/single-spa/single-spa/issues/543#issuecomment-625346522


filoxo (approved):
Nice 🙈

PR #254, opened 2020-05-12
Updated the AngularJS docs to reflect the change to the domElementGetter function

andrefrua:


internettrans (approved):
This is related to https://github.com/single-spa/single-spa-angularjs/pull/45.

Thanks for your contribution

PR #256, opened 2020-05-14
Adding a callout to canopy

TheMcMurder:
When I spoke with Canopy about open sourcing management of single-spa they were reluctant to give up the engineering credibility owning the project gave them.

I agreed that I'd add a callout to canopy where appropriate so they still get credit for creating the environment of innovation, funding the project, facilitating the vision, and dogfooding it.

I neglected to do this right away, but I'm doing it now.

Screen Shot 2020-05-14 at 9 21 15 AM

### DIFF
--- a/website/src/theme/Footer/index.js
+++ b/website/src/theme/Footer/index.js
@@ -88,17 +88,18 @@ function Footer() {
             ))}
           </div>
         )}
-        {(logo || copyright) && (
-          <div className="text--center">
-            {logo && logo.src && (
-              <img
-                className="footer__logo margin-bottom--sm"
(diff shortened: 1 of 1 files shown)


website/src/theme/Footer/index.js
```diff
@@ -88,17 +88,18 @@ function Footer() {
             ))}
           </div>
         )}
-        {(logo || copyright) && (
```
TheMcMurder, line 91: I opted to remove this and put the conditional logic around the copyright instead as logo already has conditional logic.

frehner (approved):
I'm good with this.

internettrans (approved):

filoxo (approved):
👍 🙈

PR #259, opened 2020-05-15
Fix global variable usage docs for AngularJS

internettrans:
See related https://single-spa.slack.com/archives/C8R6U7MT7/p1589540867402100


filoxo (approved):
🙈

PR #260, opened 2020-05-21
Update single-spa-angular@3 usage to include extra providers

internettrans:
See related https://github.com/single-spa/single-spa-angular/releases/tag/v3.4.0 and https://github.com/single-spa/single-spa-angular/pull/192


internettrans:
cc @arturovt who worked on the code for this.

filoxo (approved):
LGTM 🙈 but we should probably wait for @arturovt's review on this as well.

PR #268, opened 2020-06-03
Beginning documentation for single-spa-layout.

internettrans:


website/versioned_docs/version-5.x/layout-overview.md
```diff
@@ -0,0 +1,95 @@
...
+
+`single-spa-layout` is new and currently released under the `beta` dist-tag on npm. We are gathering feedback and improving the layout engine in preparation for a stable release. Although we do not expect the layout engine to change drastically, we do not recommend using it in a production setting until it is released as stable.
+
+### Browser support
+
+`single-spa-layout` works in all browsers supported by single-spa, including IE11.
+
+### Requirements
+
+You must use single-spa@>=5.4.0 in order for the layout engine to work. Additionally, you may not provide custom `domElementGetter` functions for any of your single-spa applications, as those override the configuration within single-spa-layout.
```
filoxo, line 45: Oh interesting. This would make it tricky to migrate to if an org had relied on domElementGetters for all their applications, right? One would have to find a way to remove domElementGetter, deploy updated application assets to prod but not update the prod importmap, and can then publish the updated root config and update all the imports at that time. I'm not good at visualizing this all conceptually, so please let me know if I missed something obvious. 😅

internettrans: Yes if an organization is heavily using domElementGetters they will have to migrate. Here's what a migration could look like:

Before:
```js
function domElementGetter() {
  let navEl = document.getElementById('our-nav')
  if (!navEl) {
    const navEl = document.createElement('nav');
    navEl.id = 'our-nav';
    navEl.style.position = 'fixed';
    document.body.appendChild(navEl)
  }

  return navEl
}
```

Migration / Transition period:
```js
function domElementGetter() {
  let navEl = document.getElementById('single-spa-application:our-nav') || document.getElementById('our-nav')
  if (!navEl) {
    const navEl = document.createElement('nav');
    navEl.id = 'our-nav';
    navEl.style.position = 'fixed';
    document.body.appendChild(navEl)
  }

  return navEl
}
```

After:

(Delete the domElementGetter entirely)

PR #273, opened 2020-06-05
Removes the reference to react-specific apps in the docs

thawkin3:
The docs said that `webpack-config-single-spa` was a webpack config for React single-spa applications, but I don't think that's true, right? It seems like this is a base config and then it gets extended by all the other packages like `webpack-config-single-spa-react` to make them framework-specific. I also don't see any dependencies on React in the package (https://github.com/single-spa/create-single-spa/blob/master/packages/webpack-config-single-spa/package.json), so I'm fairly certain this was a simple error in the docs.

Before:
image

After:
image


filoxo (approved):
Yep, that's a mistake. If anything the [next heading](https://single-spa.js.org/docs/create-single-spa/#webpack-config-single-spa-react) should contain that reference to React.

PR #274, opened 2020-06-05
Makes the casing consistent in many places

thawkin3:
Another petty update here. 😄 

I've corrected the casing for `html` to be `HTML` and `javascript` to be `JavaScript`.

I made sure to only change the instances in the docs that made sense and didn't touch things like `index.html` or `type="text/javascript"` where it makes sense to use all lowercase.


frehner (approved):
Whew! Thanks for updating all of that

PR #292, opened 2020-07-02
Fix link to example root config file. Resolves #291.

### LINKED ISSUE #291: "example root config" link leads to empty page
I read "Configuring single-spa" documentation page:
https://single-spa.js.org/docs/configuration/#registering-applications

And found link that leads to empty page:
|link|empty page|
|---|---|
|empty page|Screenshot 2020-07-02 at 10 24 38|

Seems like this link is used in multiple pages:
multiple pages

Should it be updated?

internettrans:
See #291


TheMcMurder (approved):
🐙

PR #293, opened 2020-07-02
Document custom webpack changes (angular)

internettrans:
See https://github.com/single-spa/single-spa-angular/pull/229#issuecomment-652748404

### DIFF
--- a/website/versioned_docs/version-5.x/ecosystem-angular.md
+++ b/website/versioned_docs/version-5.x/ecosystem-angular.md
@@ -196,6 +196,27 @@ Configuration options are provided to the `architect.serve.options` section of y
 | ---- | ----------- | ------------- |
 | singleSpaWebpackConfigPath | (optional) Path to partial webpack config to be merged with angular's config. Example: `extra-webpack.config.js` | undefined |
 
+### Use Custom Webpack
+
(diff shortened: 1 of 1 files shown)


KevinCathcart:
Looks reasonable. 

A few thoughts:
- The text at the top of the "Use Angular Builder" section should probably be updated to point users to this new section when using Angular 8 or greater.
- It may make sense to add a table of options supported in `architect.build.options.customWebpackConfig`. Of those provided by customWebPack itself, only `path` is applicable and required. (The `mergeStrategy` and `replaceDuplicatePlugins` options are not applicable since the config is exporting a function). The single-spa specific options are the same as for the old builder, except for being nested inside `customWebpackConfig`.

Thus something like: 

``` 
Configuration options are provided to the `architect.build.options.customWebpackConfig` section of your angular.json. 

| Name | Description | Default Value |
| ---- | ----------- | ------------- |
| path | (required) Path to the the above `extra-webpack.config.js` file. | N/A |
| libraryName | (optional) Specify the name of the module | Angular CLI project name |
| libraryTarget | (optional) The type of library to build [see available options](https://github.com/webpack/webpack/blob/master/declarations/WebpackOptions.d.ts#L1111) | "UMD" |
```

internettrans:
👍 updated with your feedback

PR #298, opened 2020-07-17
fix: remove .md extension in anchor tags

philtim:
Scrolling to anchor tags was not working due to having .md in the URL. By removing the file extension from the target link, scrolling works as expected. 

Keep in mind, that this only works in the production build and not in dev mode (https://github.com/facebook/docusaurus/issues/2399)


internettrans (approved):
Thanks

internettrans:
This is related to docusaurus v2 and we've dealt with it in the past. Also related to hosting on Github pages

internettrans:
This fix didn't resolve the issue. I'll submit a PR that fixes it.

PR #300, opened 2020-07-18
Fixing links

internettrans:
See related https://github.com/single-spa/single-spa.js.org/issues/297, https://github.com/single-spa/single-spa.js.org/pull/299, and https://github.com/single-spa/single-spa.js.org/issues/297.

The root of the problem here is that Docusaurus v2 doesn't really work on Github pages. You have to have a trailing slash in links for it to work. And relative links do not work as advertised. Hard coding to `/docs/` breaks how links are supposed to be relative to the version of the documentation. But we need this working for the single-spa@5 branch and I don't have the desire to figure out how to switch off of github pages today.

### DIFF
--- a/website/versioned_docs/version-5.x/building-applications.md
+++ b/website/versioned_docs/version-5.x/building-applications.md
@@ -8,7 +8,7 @@ A single-spa registered application is everything that a normal SPA is, except t
 
 ## Creating a registered application
 
-To create a registered application, first [register the application with single-spa](./configuration/#registering-applications). Once registered, the registered application must correctly implement **all** of the following lifecycle functions inside of its main entry point.
+To create a registered application, first [register the application with single-spa](/docs/configuration//#registering-applications). Once registered, the registered application must correctly implement **all** of the following lifecycle functions inside of its main entry point.
 
 ## Registered application lifecycle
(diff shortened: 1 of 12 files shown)


website/versioned_docs/version-5.x/building-applications.md
```diff
@@ -8,7 +8,7 @@ A single-spa registered application is everything that a normal SPA is, except t
 
 ## Creating a registered application
 
-To create a registered application, first [register the application with single-spa](./configuration/#registering-applications). Once registered, the registered application must correctly implement **all** of the following lifecycle functions inside of its main entry point.
+To create a registered application, first [register the application with single-spa](/docs/configuration//#registering-applications). Once registered, the registered application must correctly implement **all** of the following lifecycle functions inside of its main entry point.
```
filoxo: Possible typo: double slash after configuration

filoxo (commented):
One possible typo

website/versioned_docs/version-5.x/building-applications.md
internettrans: Yes, that's a typo. Good catch. I have updated it.

PR #303, opened 2020-07-30
Fix floating side navigation

maximillianfx:
Adding the specific css selector to fix the sidebar navigation wrong position in some pages (eg. Angular, Ember Ecosystem).


filoxo (approved):
LGTM! 🙈  Thanks!

PR #308, opened 2020-08-10
update docs for two registered applications simultaneously

jsmapr1:
Hello, 

The instructions to run two registered applications simultaneously don't seem to work. When I added a prefix to the divs, it worked as expected. Updating the docs to reflect a working example.

Here's what the change will look like.

Screen Shot 2020-08-10 at 10 17 28 AM


frehner (approved):

frehner:
thank you!

PR #312, opened 2020-08-20
Improving the definition of a Utility MFE. Also including some examples.

### LINKED ISSUE #311: Discussion: Improvement on `Utility` documentation
I've been spending some time explaining single-spa to people at Workfront and I've some potential areas for improvement in the documentation around utilities. There has been a decent amount of confusion around this point in the documentation.

Right now we [describe a `utility` as something with no routes, no single-spa api, that doesn't directly render ui, and has no lifecycles.](https://single-spa.js.org/docs/module-types)

I'm wondering if it would be better to describe it as: "An external javascript module that implements one or more aspects of a shared context outside of single-spa, but accessible to single-spa applications."

In our recommended approach a `Utility` has no direct interface with single-spa. It's just a singleton js module that exports methods that can be used by single-spa `applications`/`parcels` through `SystemJS`. A few examples of `utilities` at Canopy that would be in violation of how we describe them in the documentation:
1. `toast-service`: it directly renders ui, IIRC it isn't even initialized by the `root-config` it initializes itself when it is imported.
2. `sentry-error-logging-ui`: this has lifecycles, but the lifecycles don't match the normal single-spa lifecycles because it doesn't have it's own UI, but it does have a conceptual `bootstrap` lifecycle.
3. `feature-toggles-ui`: Before leaving canopy we added some features to this around users being able to temporarily opt-out of feature-toggles and it had ui.

I'm hoping we can have a discussion about this because I'm not sure how to describe a `utility` as an external js module without leading to more confusion.

TheMcMurder:
Closes #311

### DIFF
--- a/website/versioned_docs/version-5.x/module-types.md
+++ b/website/versioned_docs/version-5.x/module-types.md
@@ -10,13 +10,13 @@ Single-spa has [different categories](/docs/microfrontends-concept/#types-of-mic
 
 Here is how each single-spa microfrontend works conceptually. This information should help you understand our [recommendations](/docs/recommended-setup/#applications-versus-parcels-versus-utility-modules).
 
-| Topic                | Application                       | Parcel                               | Utility                              |
-| -------------------- | --------------------------------- | ------------------------------------ | ------------------------------------ |
-| Routing              | has multiple routes               | has no routes                        | has no routes                        |
-| API                  | declarative API                   | imperative API                       | no single-spa API                    |
-| Renders UI           | renders UI                        | renders UI                           | doesn't directly render UI           |
-| Lifecycles           | single-spa managed lifecycles     | custom managed lifecycles            | no lifecycles                        |
-| When to use          | core building block               | only needed with multiple frameworks | useful to share common logic         |
+| Topic                | Application                       | Parcel                               | Utility                                           |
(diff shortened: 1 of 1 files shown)


website/versioned_docs/version-5.x/module-types.md
```diff
@@ -10,13 +10,13 @@ Single-spa has [different categories](/docs/microfrontends-concept/#types-of-mic
...
 
-| Topic                | Application                       | Parcel                               | Utility                              |
-| -------------------- | --------------------------------- | ------------------------------------ | ------------------------------------ |
-| Routing              | has multiple routes               | has no routes                        | has no routes                        |
-| API                  | declarative API                   | imperative API                       | no single-spa API                    |
-| Renders UI           | renders UI                        | renders UI                           | doesn't directly render UI           |
-| Lifecycles           | single-spa managed lifecycles     | custom managed lifecycles            | no lifecycles                        |
-| When to use          | core building block               | only needed with multiple frameworks | useful to share common logic         |
+| Topic                | Application                       | Parcel                               | Utility                                           |
```
TheMcMurder, line 13: After tweaking the text in the table the table looked funny in markdown. I increased the length of the column so that it lined up nicely in markdown.

internettrans (approved):
Looks good to me

PR #316, opened 2020-08-26
Fixed typo in parcels overview

denniskigen:
Change 'single' to 'since' in the Update lifecycle explanation.


internettrans (approved):
Thanks!

PR #325, opened 2020-09-16
Update parcels-overview.md

zleight1:
Fix dead parcel links


filoxo (approved):
Thanks for catching and fixing this!

PR #326, opened 2020-09-16
Update api.md

richardbladh:
Minor change where the example code for pathToActiveWhen had the wrong path. One char added.


filoxo (approved):
Ah good catch, thanks!

PR #332, opened 2020-09-21
Updating for single-spa@5.6.0 - bootstrap is optional.

internettrans:
See related https://github.com/single-spa/single-spa/releases/tag/v5.6.0


filoxo (approved):
🙈

PR #333, opened 2020-09-22
Documentation for server-side rendering

internettrans:
This documentation is incomplete - I still need to add API documentation for single-spa-layout/server, and probably other things. But I thought I'd share here for feedback.

### DIFF
--- a/website/versioned_docs/version-5.x/server-side-rendering-overview.md
+++ b/website/versioned_docs/version-5.x/server-side-rendering-overview.md
@@ -8,27 +8,287 @@ sidebar_label: Overview
 
 In the context of single page applications (SPAs), server-side rendering (SSR) refers to dynamic generation of the HTML page that is sent from web server to browser. In a single page application, the server only generates the very first page that the user requests, leaving all subsequent pages to be rendered by the browser.
 
-To accomplish server-side rendering of an SPA, javascript code is executed in NodeJS to generate the initial HTML. In the browser, the same javascript code is executed during a "hydration" process, which attaches event listeners to the HTML. Most popular UI Frameworks (Vue, React, Angular, etc) are capable of executing in both NodeJS and the browser, and offer APIs for hydration of the application. Additionally, there are popular libraries such as NextJS and Nuxt which enhance the developer experience of server-side rendering.
+To accomplish server-side rendering of an SPA, javascript code is executed in NodeJS to generate the initial HTML. In the browser, the same javascript code is executed during a "hydration" process, which attaches event listeners to the HTML. Most popular UI Frameworks (Vue, React, Angular, etc) are capable of executing in both NodeJS and the browser, and offer APIs for both generating the server HTML and hydrating it in the browser. Additionally, there are popular frameworks such as NextJS and Nuxt which simplify the developer experience of server-side rendering.
 
 In the context of microfrontends, server-side rendering refers to assembling the HTML from multiple, separate microfrontends. Each microfrontend controls a fragment of the HTML sent from web server to browser, and hydrate their fragment once initialized in the browser.
 
 ## Purpose
 
-A primary purpose of server-side rendering is improved performance. Server-side rendered pages often display their content to users faster than their static counterparts. Other reasons for SSR include improved SEO.
+A primary purpose of server-side rendering is improved performance. Server rendered pages often display their content to users faster than their static counterparts, since the user is presented with the content before javascript resources have been initialized. Other reasons for SSR include improved search engine optimization (SEO).
+
+Server rendered applications are generally harder to build and maintain, since the code has to work on both client and server. Additionally, SSR often complicates the infrastructure needed to run your application, since many SPA + SSR solutions require NodeJS, which is not required in production for client-only SPAs.
 
 ## Implementation Overview
 
-The ultimate goal of server-side rendering is to generate an HTTP response that the browser will display to the user while javascript is hydrating. This is done with the following steps:
+The ultimate goal of server-side rendering is to generate an HTTP response that the browser will display to the user while javascript is hydrating. Most microfrontend server-side rendering implementations, including single-spa's recommended approach, do this with the following steps:
 
-1. Identify which microfrontends to render for the incoming HTTP request.
-2. Render the HTML for each microfrontend to an
+1. Layout - Identify which microfrontends to render for the incoming HTTP request, and where within the HTML they will be placed. This is usually route based.
+2. Fetch - Begin rendering the HTML for each microfrontend to a stream.
+3. Headers - Retrieve HTTP response headers from each microfrontend. Merge them together and send the result as the HTTP response headers to the browser.
+4. Body - Send the HTTP response body to the browser, which is an HTML document consisting of static and dynamic parts. This involves waiting for each microfrontend's stream to end before proceeding to the next portion of HTML.
+5. Hydrate - Within the browser, download all javascript needed to hydrate the HTML.
 
-## Defining Layout
+## 1. Layout
 
-## HTTP Response Headers
+To define an HTML template that lays out your page, first choose a "microfrontend layout middleware":
 
-## HTTP Response Body
+1. [single-spa-layout](/docs/layout-overview): The official layout engine for single-spa.
+2. [Tailor](https://github.com/zalando/tailor): A popular, battle tested layout engine that predates single-spa-layout and is not officially affiliated with single-spa.
+3. [TailorX](https://github.com/StyleT/tailorx): An actively maintained fork of Tailor that is used by [Namecheap](https://www.namecheap.com/) in their single-spa website. The single-spa core team collaborated with the creators of TailorX when authoring single-spa-layout, taking some inspiration from it.
 
-### Shared server with dynamic module loading
+We generally recommend single-spa-layout, although choosing one of the other options might make sense for your situation, since single-spa-layout is newer and less battle tested than Tailor/TailorX.
 
-### Separate servers over HTTP
+With single-spa-layout, you define a single template that handles all routes. [Full documentation](/docs/layout-definition).
+
+```html
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <meta http-equiv="X-UA-Compatible" content="ie=edge" />
+    <title>Isomorphic Microfrontends</title>
+    <meta
+      name="importmap-type"
+      content="systemjs-importmap"
+      server-cookie
+      server-only
+    />
+    <script src="https://cdn.jsdelivr.net/npm/import-map-overrides@2.0.0/dist/import-map-overrides.js"></script>
+    <script src="https://cdn.jsdelivr.net/npm/systemjs@6.6.1/dist/system.min.js"></script>
+    <script src="https://cdn.jsdelivr.net/npm/systemjs@6.6.1/dist/extras/amd.min.js"></script>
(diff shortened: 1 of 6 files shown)


website/versioned_docs/version-5.x/server-side-rendering-overview.md
```diff
@@ -0,0 +1,294 @@
...
+id: ssr-overview
+title: Server Side Rendering
+sidebar_label: Overview
+---
+
+## Intro to SSR
+
+In the context of single page applications (SPAs), server-side rendering (SSR) refers to dynamic generation of the HTML page that is sent from web server to browser. In a single page application, the server only generates the very first page that the user requests, leaving all subsequent pages to be rendered by the browser.
+
+To accomplish server-side rendering of an SPA, javascript code is executed in NodeJS to generate the initial HTML. In the browser, the same javascript code is executed during a "hydration" process, which attaches event listeners to the HTML. Most popular UI Frameworks (Vue, React, Angular, etc) are capable of executing in both NodeJS and the browser, and offer APIs for both generating the server HTML and hydrating it in the browser. Additionally, there are popular frameworks such as NextJS and Nuxt which simplify the developer experience of server-side rendering.
```
TheMcMurder, line 11: > Additionally, there are popular frameworks such as NextJS and Nuxt which simplify the developer experience of server-side rendering.

Reading this seems to indicate that single-spa server rendering supports NextJS and Nuxt. This paragraph never explicitly states that it does, I wonder if including this will confuse people.

TheMcMurder (approved):

TheMcMurder:
This seems like a really good overview of how SSR works in an MFE context and a intro into the different options available for single-spa + SSR.

Love it.

website/versioned_docs/version-5.x/server-side-rendering-overview.md
internettrans: I hope to add support for Next / Nuxt over time.

website/versioned_docs/version-5.x/server-side-rendering-overview.md
```diff
@@ -0,0 +1,294 @@
...
+
+## 1. Layout
+
+To define an HTML template that lays out your page, first choose a "microfrontend layout middleware":
+
+1. [single-spa-layout](/docs/layout-overview): The official layout engine for single-spa.
+2. [Tailor](https://github.com/zalando/tailor): A popular, battle tested layout engine that predates single-spa-layout and is not officially affiliated with single-spa.
+3. [TailorX](https://github.com/StyleT/tailorx): An actively maintained fork of Tailor that is used by [Namecheap](https://www.namecheap.com/) in their single-spa website. The single-spa core team collaborated with the creators of TailorX when authoring single-spa-layout, taking some inspiration from it.
+
+We generally recommend single-spa-layout, although choosing one of the other options might make sense for your situation, since single-spa-layout is newer and less battle tested than Tailor/TailorX.
```
filoxo: I feel like the phrase "less battle tested" could cause people to lose confidence in the library before even trying it.

```suggestion
We generally recommend single-spa-layout, although choosing one of the other options might make sense for your situation, since single-spa-layout is newer and thus less mature than Tailor/TailorX.
```

website/versioned_docs/version-5.x/server-side-rendering-overview.md
```diff
@@ -0,0 +1,294 @@
...
+
+With single-spa-layout, you define a single template that handles all routes. [Full documentation](/docs/layout-definition).
```
filoxo, line 41: Hopefully this isn't too picky but more explicit wording of links is better for a11y and SEO.

```suggestion
With single-spa-layout, you define a single template that handles all routes. [Refer to the full Layout Definition documentation here](/docs/layout-definition).
```

website/versioned_docs/version-5.x/server-side-rendering-overview.md
```diff
@@ -0,0 +1,294 @@
...
+      server-cookie
+      server-only
+    />
+    <script src="https://cdn.jsdelivr.net/npm/import-map-overrides@2.0.0/dist/import-map-overrides.js"></script>
+    <script src="https://cdn.jsdelivr.net/npm/systemjs@6.6.1/dist/system.min.js"></script>
+    <script src="https://cdn.jsdelivr.net/npm/systemjs@6.6.1/dist/extras/amd.min.js"></script>
+    <script src="https://cdn.jsdelivr.net/npm/systemjs@6.6.1/dist/extras/named-exports.min.js"></script>
+  </head>
+  <body>
+    <template id="single-spa-layout">
```
filoxo, line 63: The [Layout Definition](https://single-spa.js.org/docs/layout-definition/#html-layouts) docs put the `template` element in the `head`. I vaguely recall that that is to prevent it from being rendered, but with SSR that doesn't matter because its being pre-rendered (if I'm not mistaken... which I'm not confident I am). Is that worth noting somewhere? Might be better to put over in that page instead though... 🤔

website/versioned_docs/version-5.x/server-side-rendering-overview.md
```diff
@@ -0,0 +1,294 @@
...
+
+  bodyStream.pipe(res);
+}).listen(9000)
+```
+
+## 3. HTTP Response Headers
+
+The HTTP response headers sent to the browser are a combination of default headers and the headers retrieved from each microfrontend. Your [method of fetching microfrontends](#2-fetch) does not change how the final headers are merged and assembled for the browser.
+
+Tailor and TailorX have default methods of merging headers. At the time of this writing, single-spa-layout requires manual merging of the headers.
```
filoxo: This seems to be more of an aside so it might be good to put it into a side comment (though Blockquotes shouldn't be used for this... but that's how they're used almost everywhere with Markdown).

```suggestion
> At the time of this writing, single-spa-layout requires manual merging of the headers; while Tailor and TailorX have default methods of merging headers. 
```

website/versioned_docs/version-5.x/server-side-rendering-overview.md
```diff
@@ -0,0 +1,294 @@
...
+    bodyStream.pipe(res);
+  })
+}).listen(9000)
+```
+
+## 5. Hydrate
+
+[Helpful article about hydration](https://developers.google.com/web/updates/2019/02/rendering-on-the-web)
+
+Hydration (or rehydration) refers to browser Javascript initializing and attaching event listeners to the HTML sent by the server. There are several variants, including progressive rehydration and partial rehydration.
```
filoxo: I think this one line description is very good, so I'd put the link below it and as an aside.

```suggestion
Hydration (or rehydration) refers to browser Javascript initializing and attaching event listeners to the HTML sent by the server. There are several variants, including progressive rehydration and partial rehydration.

> See also ["Rendering on the Web"](https://developers.google.com/web/updates/2019/02/rendering-on-the-web) by Google.
```

website/versioned_docs/version-5.x/server-side-rendering-overview.md
```diff
@@ -0,0 +1,294 @@
...
+
+In the context of microfrontends, hydration is done by the underlying UI framework of the microfrontend (React, Vue, Angular, etc). For example, in React, this is done by calling [ReactDOM.hydrate()](https://reactjs.org/docs/react-dom.html#hydrate). The [single-spa adapter libraries](http://localhost:3000/docs/ecosystem) allow you to specify whether you are hydrating or mounting for the first time (see single-spa-react's [`renderType` option](/docs/ecosystem-react#options)).
```
filoxo:
```suggestion
In the context of microfrontends, hydration is done by the underlying UI framework of the microfrontend (React, Vue, Angular, etc). For example, in React, this is done by calling [ReactDOM.hydrate()](https://reactjs.org/docs/react-dom.html#hydrate). The [single-spa adapter libraries](/docs/ecosystem) allow you to specify whether you are hydrating or mounting for the first time (see single-spa-react's [`renderType` option](/docs/ecosystem-react#options)).
```

filoxo (commented):
I left a few comments. They're mostly low priority.

website/versioned_docs/version-5.x/server-side-rendering-overview.md
internettrans: Perhaps they should lose confidence in it 😄. It has never been used for anything serious

website/versioned_docs/version-5.x/server-side-rendering-overview.md
internettrans: Fully self contained link text is a new accessibility concern for me to think about. Doing so would require changing hundreds of links across the documentation site, and I do not see the benefits. For both visual users and screen reader users, the text surrounding a link is crucial to understanding what the link is for and why it exists. Self contained link text makes the content read awkwardly in many cases, including this one.

I am not planning on making these changes - if I'm missing a reason why this is important, feel free to share more.

website/versioned_docs/version-5.x/server-side-rendering-overview.md
internettrans: The content of `<template>` elements is never rendered.

website/versioned_docs/version-5.x/server-side-rendering-overview.md
internettrans: I don't understand what this change does.

website/versioned_docs/version-5.x/server-side-rendering-overview.md
filoxo: >  Doing so would require changing hundreds of links across the documentation site

My suggestion was for just this one line, I don't think its necessary to jump to the extreme of having to change it for hundreds of links in one go. All others can be changed over time.

> For both visual users and screen reader users, the text surrounding a link is crucial to understanding what the link is for and why it exists. 

This is not always true, and most accessibility standards will disagree with you on this. For example: 

https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-refs.html
> The intent of this Success Criterion is to help users understand the purpose of each link so they can decide whether they want to follow the link. Whenever possible, provide link text that identifies the purpose of the link without needing additional context. Assistive technology has the ability to provide users with a list of links that are on the Web page. Link text that is as meaningful as possible will aid users who want to choose from this list of links. Meaningful link text also helps those who wish to tab from link to link. 

https://webaim.org/techniques/hypertext/
> Implication 1: Links should make sense out of context. Phrases such as "click here," "more," "click for details," and so on are ambiguous when read out of context. 

https://usability.yale.edu/web-accessibility/articles/links#disabilities
> While screen readers can read a full page to a user, screen reader users may prefer to instead listen to a list of links. In that case, a screen reader may only read the link text and not the surrounding text.

But do what you will. I tried to approach this as "not a big deal".

website/versioned_docs/version-5.x/server-side-rendering-overview.md
filoxo: [Template elements are rendered in IE](https://stackoverflow.com/a/45680787/2554793), but of course that's on the client. I'm just carrying over what I remember from that example code since it might be a confusing point.

website/versioned_docs/version-5.x/server-side-rendering-overview.md
internettrans: >  I tried to approach this as "not a big deal".

Sounds like my previous comment maybe felt aggressive to you. I think this pull request is a big deal for me because it is the culmination of ~6 months of (on and off) work on server-side rendering. I was hoping more feedback from more reviewers, and feedback related to the substantive content. I think I got a bit frustrated seeing that there weren't many comments talking about the substance of the documentation here. This isn't your fault (as you gave some good comments in your review, and took the time to review it at all), and I should have reframed what I said about accessibility here. I appreciate your focus on accessibility.

PR #340, opened 2020-10-02
Updating single-spa-layout now that it's out of beta.

internettrans:
single-spa-layout@1.0.1 is released, which is no longer under the `beta` dist-tag. See https://github.com/single-spa/single-spa-layout/releases/tag/v1.0.1

### DIFF
--- a/website/versioned_sidebars/version-5.x-sidebars.json
+++ b/website/versioned_sidebars/version-5.x-sidebars.json
@@ -98,7 +98,7 @@
     },
     {
       "type": "category",
-      "label": "Layout Engine (Beta)",
+      "label": "Layout Engine",
       "items": [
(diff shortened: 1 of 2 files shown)


website/versioned_sidebars/version-5.x-sidebars.json
```diff
@@ -210,7 +210,7 @@
     },
     {
       "type": "category",
-      "label": "Server Side Rendering (Beta)",
+      "label": "Server Side Rendering (New)",
```
internettrans, line 213: The word Beta deterred a lot of people from using Layout Engine. Although this is somewhat intentional, I think that the word beta makes people more nervous than they need to be when trying out the layout engine and ssr

PR #341, opened 2020-10-02
Deployment previews on pull requests

jasonbarry:
I think this repo would benefit in having deployment previews on every pull request, that way we can generate an automatic build of the website per PR rather than having to checkout branches locally to review/verify changes.

These are the changes needed to integrate w/ [FeaturePeek](https://featurepeek.com), a deployment preview provider. If the GitHub App is installed (for webhooks), then every subsequent PR will have its own deployment preview.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,9 +27,10 @@ jobs:
             cd website 
             yarn install --frozen-lockfile
             yarn build
-      - run:
-          name: Ping FeaturePeek
-          command: bash <(curl -s https://peek.run/ci)
+            if [ "$CIRCLE_BRANCH" != "master" ]; then
+              # Ping FeaturePeek
+              bash <(curl -s https://peek.run/ci)
+            fi
 workflows:
   version: 2
   build_and_deploy:
@@ -45,7 +46,6 @@ workflows:
             branches:
               ignore:
                 - gh-pages
-                - master
       - deploy-website:
           filters:
             branches:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,6 +27,9 @@ jobs:
             cd website 
             yarn install --frozen-lockfile
             yarn build
+      - run:
+          name: Ping FeaturePeek
+          command: bash <(curl -s https://peek.run/ci)
 workflows:
   version: 2
   build_and_deploy:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -27,6 +27,9 @@ jobs:
             cd website 
             yarn install --frozen-lockfile
             yarn build
+      - run:
+          name: Ping FeaturePeek
+          command: bash <(curl -s https://peek.run/ci)
```
internettrans: Would probably be best to only run this on non-master branches, since the master branch is going to be deployed to the live site.

internettrans (commented):
I like it, thanks. Could you share an example of a preview site on peek.run? How will we know the link to the peek.run preview for each pull request?

jasonbarry:
@joeldenning here's an example: https://peek.run/902hn27j 

When the [GitHub App](https://github.com/apps/featurepeek) is installed, the FeaturePeek bot will post a comment that contains the link.

internettrans:
I installed the Github app for this repository. Once the previews are only deployed for non-master branches, I'll approve and merge.

.circleci/config.yml
jasonbarry: @joeldenning I've added a line to ignore the master branch on the `build-website` job – please double-check that that's correct.

.circleci/config.yml
```diff
@@ -42,6 +45,7 @@ workflows:
             branches:
               ignore:
                 - gh-pages
+                - master
```
internettrans: I think this will break things, since we need the build to run on the master branch. It might make sense to make a new job called `deploy-preview` and then do the branch filter on that job instead of on the `build-website` job

jasonbarry: @joeldenning take a look at the latest diff – I put a condition inside the existing `build-website` job (inspired by https://circleci.com/docs/2.0/configuration-reference/#ending-a-job-from-within-a-step)

internettrans (approved):
Looks good, thanks. I think that circle ci is only configured to run on branches on the main repository, not forks. So we might need to fix that, too. I am going to test the peek preview with a branch right after merging this pull request

jasonbarry:
@joeldenning cool! If you create a free account on https://dashboard.featurepeek.com, you can claim the `single-spa` team since you're in the org. Let me know how it goes!

internettrans:
I went to create an account with feature peek, but it asked for more permissions than I am comfortable with.

image

Is it possible to use FeaturePeek without giving full write access to the repository and its settings?

PR #342, opened 2020-10-02
Test peek preview

internettrans:
@jasonbarry this is the pull reuqest I'm using to test the preview feature

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,6 +29,7 @@ jobs:
             yarn build
             if [ "$CIRCLE_BRANCH" != "master" ]; then
               # Ping FeaturePeek
+              cd ..
               bash <(curl -s https://peek.run/ci)
             fi
 workflows:
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 .DS_Store
+
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,1 @@
 .DS_Store
-


internettrans:
The preview failed to deploy with the following error:

```
Integration not found
Failed to send ping to FeaturePeek API.
```

 https://app.circleci.com/pipelines/github/single-spa/single-spa.js.org/420/workflows/a4daccff-b003-4b47-9ee8-0d754fc5b281/jobs/1175.

However, the FeaturePeek Github app is installed. Do you know what's wrong, @jasonbarry?

image

jasonbarry:
@joeldenning ah, there should be a `cd ..` step right before the bash script gets curled in CI. Since the first line in the command is `cd website`, we need to reset that. (Before, it was in a separate `run` step, which auto-reset the cwd.) 

Also, you may need to make a free FeaturePeek [account](https://dashboard.featurepeek.com) and claim your team to complete the integration.

internettrans:
> here should be a cd .. step right before the bash script gets curled in CI

Ah makes sense I'll try that

> Also, you may need to make a free FeaturePeek account and claim your team to complete the integration.

I'm not comfortable with this, since FeaturePeek asked for significantly more permissions than it should need when I tried to sign up with github.

internettrans:
Looks like the `cd ..` didn't fix it.

https://app.circleci.com/pipelines/github/single-spa/single-spa.js.org/421/workflows/a2a0ea30-6f45-4a0e-a835-c085e826159c/jobs/1176

jasonbarry:
@joeldenning `cd ..` looks good, as now it finds that it's a static build (`Static build detected` is new this time)

Regarding OAuth permissions, we try to follow the principle of least privilege. Our API assigns the deployment to a team, but in this case the team hasn't been created yet on FeaturePeek, which is why it's failing. I'll see if we can auto-create the team when the GH app is installed – usually it gets created during user onboarding.

internettrans:
Peek should need zero access to do what we're trying for in this case, since viewing code and commenting on public Github requires no permissions. Instead, it asked me for admin-level access to all of my repository settings, secrets, and code.

internettrans:
If I sign up with my email address instead of via Github, will I be able to claim the organization? Or would I still need to link my Github in order to verify ownership of the single-spa Github org?

jasonbarry:
Thanks for your feedback – we use the same OAuth App for user login for public and private repos, which is why we ask for that even if you only intend to use it on public repos. You can also choose for the GitHub App to only have access to select repositories vs all repositories. 

You can sign up with email/password, but the next step would be to link your GitHub, which would ask for the same permissions.

internettrans:
Sounds good. For now, I will revert the changes in the previous pull request and close this pull request. If the organization is claimed such that previews would start working, I'm happy to reinstate this.

PR #343, opened 2020-10-08
Update ecosystem-angular.md

lee9019:
Added parcel component usage in angular

### DIFF
--- a/website/versioned_docs/version-5.x/ecosystem-angular.md
+++ b/website/versioned_docs/version-5.x/ecosystem-angular.md
@@ -781,3 +781,33 @@ const lifecycles = singleSpaAngular({
 ```
 
 > ⚠️ `single-spa-angular@4.x` requires calling `getSingleSpaExtraProviders` function in applications that have routing. Do not call this function in _zone-less_ application.
+
+### Parcels
+
+In order to consume parcels created in other frameworks (vue, react) in angular, there is a angular parcel component available which can be used to fetch the vue or react parcel and add it to the DOM in angular. The link to the parcel component https://github.com/ricardofct/single-spa-angular/blob/angular-parcel/src/browser-lib/parcel-lib/parcel.component.ts. Copy this component and paste it in your angular code.
+
+To use the component in html
+```html
+<parcel *ngIf="spaSite" [config]="spaSite" [mountParcel]="mountParcel"> </parcel>
+```
+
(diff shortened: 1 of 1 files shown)


lee9019:
@joeldenning Could you check this?

website/versioned_docs/version-5.x/ecosystem-angular.md
```diff
@@ -781,3 +781,33 @@ const lifecycles = singleSpaAngular({
 ```
 
 > ⚠️ `single-spa-angular@4.x` requires calling `getSingleSpaExtraProviders` function in applications that have routing. Do not call this function in _zone-less_ application.
+
+### Parcels
+
+In order to consume parcels created in other frameworks (vue, react) in angular, there is a angular parcel component available which can be used to fetch the vue or react parcel and add it to the DOM in angular. The link to the parcel component https://github.com/ricardofct/single-spa-angular/blob/angular-parcel/src/browser-lib/parcel-lib/parcel.component.ts. Copy this component and paste it in your angular code.
```
filoxo, line 787: This isn't correct. Parcels have been added to single-spa-angular's public API, see here: https://github.com/single-spa/single-spa-angular/blob/master/src/src/public_api.ts#L3-L4
You should instead be able to `import { ParcelModule, ParcelComponent } from 'single-spa-angular'` but I don't have an Angular app to validate that at the moment. Will you try out that syntax and update these docs?

filoxo (changes requested):
Thanks for contributing! I left one comment about how to use the Parcel component. Can I also ask you to try to follow the same format and outline as [single-spa-react's Parcel docs](https://single-spa.js.org/docs/ecosystem-react/#parcels)?

website/versioned_docs/version-5.x/ecosystem-angular.md
internettrans: Agreed. I don't think that that import works ever since we switched to ng-packagr instead of rollup for bundling single-spa-angular. We should not instruct users to copy/paste code, but rather to use it from their node_modules.

internettrans:
The parcel component has been fixed, and there is a working example in https://github.com/single-spa/single-spa-angular/releases/tag/v4.7.0. @arturovt is planning on submitting a PR with documentation for it. I am closing this in favor of that PR

PR #345, opened 2020-10-12
Adding documentation for canceling navigation

internettrans:


website/versioned_docs/version-5.x/api.md
```diff
@@ -655,17 +655,41 @@ window.addEventListener('popstate', evt => {
 });
 ```
 
+### Canceling navigation
+
+Canceling navigation refers to the URL changing and then immediately changing back to what it was before. This is done before any mounting, unmounting, or loading that would otherwise take place. . This can be used in conjunction with Vue router and Angular router's built-in navigation guards that allow for cancelation of a navigation event.
```
filoxo:
```suggestion
Canceling navigation refers to the URL changing and then immediately changing back to what it was before. This is done before any mounting, unmounting, or loading that would otherwise take place. This can be used in conjunction with Vue router and Angular router's built-in navigation guards that allow for cancelation of a navigation event.
```

filoxo (approved):
One minor change, but overall LGTM 🙈

PR #346, opened 2020-10-13
Update getting started guide to use create-single-spa

### LINKED ISSUE #335: Update "Getting started" with create-single-spa
The ["Getting Started"](https://single-spa.js.org/docs/getting-started-overview) page should have a section that guides a user to:

1. create a root config
2. create a basic application (generic? or perhaps starting with single-spa-html?)
3. wiring up that application to the root-config

single-spa-playground kind of serves this purpose but I've helped a few people who do not understand what to do once their application is running there. 

Opening this discussion here to get feedback on whether this should be a beginner-to-single-spa's starting point.

filoxo:
Resolves #335 by updating the "Quick Start" guide with instructions on using create-single-spa to generate boilerplate for the root-config and a first single-spa application. I think that section grew large enough that it *might* be large enough to put into its own document but wanted to get some feedback before doing that. Some of the changes are just Prettier's formatting changes.

### DIFF
--- a/website/versioned_docs/version-5.x/getting-started-overview.md
+++ b/website/versioned_docs/version-5.x/getting-started-overview.md
@@ -22,18 +22,18 @@ See [our examples page](/docs/examples).
 
 ## Architectural Overview
 
-single-spa takes inspiration from modern framework component lifecycles by applying lifecycles to entire applications.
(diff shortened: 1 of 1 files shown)


internettrans (approved):
Very good updates

PR #351, opened 2020-10-26
Clarify that route paths are prefixes.

internettrans:
See related https://github.com/single-spa/single-spa-layout/issues/85


filoxo (approved):
🙈

PR #373, opened 2020-11-28
Update docs to reflect single-spa-html PR.

thantos:
Update docs to reflect html props PR.

https://github.com/single-spa/single-spa-html/pull/14


internettrans:
Thanks for also updating the docs here 👍

internettrans (approved):

PR #374, opened 2020-11-28
Update single-spa-preact docs.

thantos:
Single-spa-preact domElementGetter receives single-spa props as of 8/2020. https://github.com/single-spa/single-spa-preact/blob/master/src/single-spa-preact.js#L77


internettrans (approved):
Thanks

PR #382, opened 2020-12-16
Fix link of custom props

chinesedfan:
Make it the same with https://single-spa.js.org/docs/ecosystem-vue#custom-props-1.

And in fact, the website has a typo but accessible link, https://single-spa.js.org/docs/building-applications/#lifecyle-props.


filoxo (approved):
Thanks for contributing this fix!

PR #383, opened 2020-12-17
fixed typo

kroissma:


frehner:
Thanks for the PR.

However, I'm not entirely sure this is correct; I read it as 

>each application has its own loaders

and not as 

>there are loaders that run before applications

while I do believe both are true, I think the original intent was the first statement. 

It may be useful to just somewhat rewrite the sentence if it's confusing though.

internettrans:
`per-application` is intended. I agree with @frehner that perhaps the whole sentence could be reworked. But `per-application` is not a typo.

PR #384, opened 2020-12-23
Adding standalone mode to local development docs.

internettrans:
See https://github.com/single-spa/standalone-single-spa-webpack-plugin/issues/8#issuecomment-750097740


hogarthww-labs (approved):
Great :)

PR #390, opened 2021-01-14
Upgrade to latest Docusaurus

filoxo:
- upgrade docusaurus from 2.0.0-alpha.37 to 2.0.0-alpha.70
- renamed custom redirect plugin to silence a warning about official docusaurus plugins
- updated block quotes with [Callouts](https://v2.docusaurus.io/docs/markdown-features/#calloutsadmonitions)
- updated code block filenames with [Code block Titles](https://v2.docusaurus.io/docs/markdown-features/#code-title)
- fixed various links that docusaurus was warning about
- fix CSS for line highlighting
- applied prettier formatting to markdown and code blocks

### DIFF
--- a/website/blog/2020-02-24-single-spa-5.md
+++ b/website/blog/2020-02-24-single-spa-5.md
@@ -37,7 +37,7 @@ The ESM version of single-spa@4 was 23.8kb (7.2kb gzipped). That was improved in
 
 Since single-spa's inception, bundler configuration has been a huge source of user pain. We have heard this pain and implemented [create-single-spa](/docs/create-single-spa), which creates (and sometimes can update) repositories that are ready to be used as single-spa microfrontends. For Angular and Vue, the official CLIs are used with a few extra plugins automatically installed. For React, a default webpack config with decent eslint / prettier defaults is set up.
 
-Additionally, we have added a lot of documentation for webpack in [The Recommended Setup](/docs/next/recommended-setup#build-tools-webpack--rollup).
+Additionally, we have added a lot of documentation for webpack in [The Recommended Setup](/docs/recommended-setup#build-tools-webpack--rollup).
 
 ## Tutorial videos
 
--- a/website/src/plugins/docusaurus-plugin-redirects/src/index.js
+++ b/website/src/plugins/docusaurus-plugin-redirects/src/index.js
@@ -9,7 +9,7 @@ module.exports = function(context, opts) {
   const options = { ...DEFAULT_OPTIONS, ...opts };
 
   return {
-    name: 'docusaurus-plugin-redirects',
+    name: 'custom-docusaurus-plugin-redirects',
 
     async postBuild({ siteConfig = {}, routesPaths = [], outDir }) {
       routesPaths.map(routesPath => {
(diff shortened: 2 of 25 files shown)


website/blog/2020-02-24-single-spa-5.md
```diff
@@ -92,4 +92,4 @@ This change does not mean anything drastic for single-spa. Its license was and i
...
 
-Please [contribute to our code](http://localhost:3000/docs/next/contributing-overview) and [ecosystem](/docs/ecosystem), [join our single-spa slack channel](https://join.slack.com/t/single-spa/shared_invite/zt-jy1x8x93-52Tth0fZBUzZmHcEpWw2Jg), [follow our official Twitter account](https://twitter.com/Single_spa), and contribute to [our open collective](https://opencollective.com/single-spa). The [single-spa core team](/contributors) all have full-time jobs and maintain this project on a volunteer basis.
\ No newline at end of file
+Please [contribute to our code](http://localhost:3000/docs/next/contributing-overview) and [ecosystem](/docs/ecosystem), [join our single-spa slack channel](https://join.slack.com/t/single-spa/shared_invite/zt-jy1x8x93-52Tth0fZBUzZmHcEpWw2Jg), [follow our official Twitter account](https://twitter.com/Single_spa), and contribute to [our open collective](https://opencollective.com/single-spa). The [single-spa core team](/contributors) all have full-time jobs and maintain this project on a volunteer basis.
```
filoxo, line 95: I just noticed this reference to localhost...

website/src/plugins/docusaurus-plugin-redirects/src/index.js
```diff
@@ -31,18 +31,13 @@ module.exports = function(context, opts) {
...
-        window.location.assign(redirectLink)
-      })
-    </script>
+    <meta http-equiv="refresh" content="0; url=${newLink}">
+    <link rel="canonical" href="${newLink}" />
+    <title>Redirecting to ${newLink}</title>
+  </head>
+  <script>
+    window.location.href = '${newLink}';
+  </script>
```
filoxo, line 40: The template here is a hybrid of the old one and the one provided by [@docusaurus/plugin-client-redirects](https://v2.docusaurus.io/docs/api/plugins/@docusaurus/plugin-client-redirects/) ([see the template here](https://github.com/facebook/docusaurus/blob/master/packages/docusaurus-plugin-client-redirects/src/templates/redirectPage.template.html.ts)). The reason we can't use that instead is because it does not generate files the same way. The new plugin creates `index.html` nested within folders, so instead of generating a `replacement.html` page, it would create `replacement.html/index.html` which isn't what we want.

website/static/css/custom.css
```diff
@@ -195,6 +181,10 @@ html[data-theme='dark'] .args-list > dt {
   content: ': ';
 }
 
+.dl-inline dd {
+  margin: 0;
+}
```
filoxo, line 186: Forgot to note this in the summary. This style is to fix this 👇🏼 
Screen Shot 2021-01-14 at 9 54 04 AM

frehner (approved):

website/blog/2020-02-24-single-spa-5.md
filoxo: I'll fix this in a follow up PR.

PR #397, opened 2021-01-21
Add Vue public path instructions

internettrans:
See related https://github.com/joeldenning/systemjs-webpack-interop/issues/21. This was also encountered at Fitch when I was consulting. So I figured it would be good to document, since it's happening fairly commonly.


filoxo (approved):
🙈

PR #399, opened 2021-01-26
Documenting AngularJS parcels

internettrans:
See related https://github.com/single-spa/single-spa-angularjs/pull/55 and https://github.com/single-spa/single-spa-angularjs/issues/46


filoxo (approved):
🙈

PR #401, opened 2021-02-02
Snowpack documentation

internettrans:
This documentation was heavily based on the Vite documentation, as snowpack and vite are fairly similar in their local development workflow.

### DIFF
--- a/website/versioned_docs/version-5.x/ecosystem-snowpack.md
+++ b/website/versioned_docs/version-5.x/ecosystem-snowpack.md
@@ -92,7 +92,9 @@ module.exports = {
 
 Snowpack works well with [development via import map overrides](https://single-spa.js.org/docs/recommended-setup#local-development). You should use http://localhost:8080/index.js as the URL for your import map override.
 
+:::caution
 Static Assets currently do not load from the correct URL in development, pending a PR to Snowpack: https://github.com/snowpackjs/snowpack/pull/2407. However, static assets do load from the correct URL in production, due to systemjs-webpack-interop.
+:::
 
 ## Native Modules vs SystemJS
 
--- a/website/versioned_docs/version-5.x/ecosystem-snowpack.md
+++ b/website/versioned_docs/version-5.x/ecosystem-snowpack.md
@@ -92,7 +92,7 @@ module.exports = {
 
 Snowpack works well with [development via import map overrides](https://single-spa.js.org/docs/recommended-setup#local-development). You should use http://localhost:8080/index.js as the URL for your import map override.
 
-Static Assets currently do not load from the correct URL, pending a PR to Snowpack: https://github.com/snowpackjs/snowpack/pull/2407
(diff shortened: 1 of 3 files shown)


website/versioned_docs/version-5.x/ecosystem-snowpack.md
```diff
@@ -0,0 +1,102 @@
...
+          delete config.optimization.runtimeChunk;
+          delete config.optimization.splitChunks;
+
+          return merge(config, {
+            mode: "development",
+            module: {
+              rules: [
+                {
+                  parser: {
+                    system: false,
```
filoxo, line 56: Would it be worth noting that this is only compatible with whatever versions of `@snowpack/plugin-webpack` use Webpack 4?

website/versioned_docs/version-5.x/ecosystem-snowpack.md
```diff
@@ -0,0 +1,102 @@
...
+    baseUrl: "http://localhost:8080/",
+  },
+};
+```
+
+## Local development
+
+Snowpack works well with [development via import map overrides](https://single-spa.js.org/docs/recommended-setup#local-development). You should use http://localhost:8080/index.js as the URL for your import map override.
+
+Static Assets currently do not load from the correct URL in development, pending a PR to Snowpack: https://github.com/snowpackjs/snowpack/pull/2407. However, static assets do load from the correct URL in production, due to systemjs-webpack-interop.
```
filoxo, line 98: This could be made into an info or warning callout using this syntax https://v2.docusaurus.io/docs/markdown-features/#calloutsadmonitions

website/versioned_docs/version-5.x/ecosystem-snowpack.md
internettrans: There are no versions of `@snowpack/plugin-webpack` that use webpack 5 yet, as far as I know. See https://github.com/snowpackjs/snowpack/blob/41c29f25e53695ec84c07386af7eb140d0b22a8e/plugins/plugin-webpack/package.json#L28

filoxo (approved):
This Snowpack example is pretty cool! Other than those minor comments, everything else looks great to me.

website/versioned_docs/version-5.x/ecosystem-snowpack.md
filoxo: Right, I'm just thinking of if/when Snowpack does publish a webpack@5 version, it might trip people up since this config wouldn't be compatible. But maybe we can add/update once such a version does exist.

website/versioned_docs/version-5.x/ecosystem-snowpack.md
internettrans: 👍  updated

website/versioned_docs/version-5.x/ecosystem-snowpack.md
internettrans: I'll put a comment in now

PR #406, opened 2021-02-08
docs(setup): clarify recommendation for shared-deps

jLouzado:
previous statement could be mis-construed, the edit attempts to make it clear that there's only one type of in-browser module being talked about.

key change is the inclusion of the word `also`, everything else was just supporting that change.

### DIFF
--- a/website/versioned_docs/version-5.x/recommended-setup.md
+++ b/website/versioned_docs/version-5.x/recommended-setup.md
@@ -34,7 +34,7 @@ is via [webpack externals](https://webpack.js.org/configuration/externals/#root)
 Here are our recommendations:
 
 1. Each single-spa application should be an in-browser Javascript module.
-2. Large shared dependencies (ie, the react, vue, or angular libraries) should each be in-browser modules.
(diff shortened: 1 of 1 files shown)


internettrans (commented):
> previous statement could be mis-construed, the edit attempts to make it clear that there's only one type of in-browser module being talked about.

I honestly don't understand how the language change clarifies the meaning here, nor do I understand what your original confusion was about. Could you help me understand how the existing language could be misinterpreted?

jLouzado:
> Could you help me understand how the existing language could be misinterpreted?

The change is intended to make the sentence more beginner-friendly, by just reiterating that each SPA is getting the same treatment as each large-dependancy.

When I first read it, it seemed as if something different was required for the two categories. The first line specifies `javascript module` and the second line just says `module`. Obviously the second line is referring to a javascript module but if someone is reading the guide while they're still learning all the new concepts there might be scope for confusion.

Rather than repeating `js module` in the second line it seemed to flow better to specify that large dependencies are "also" used as modules.

internettrans (approved):
Ah I see, thanks for the clarification. I agree - thanks for your contribution.

PR #407, opened 2021-02-08
Clarify redirect url requirements.

internettrans:
See https://github.com/single-spa/single-spa-layout/issues/116 which points out that leading slashes are okay.


filoxo (approved):
🙈

PR #408, opened 2021-02-08
Remove project status warning in angular docs

internettrans:
See https://single-spa.slack.com/archives/CGETM8T5X/p1612532310102400. Since we added this notice, arturo and danil have stepped into the role of maintainers.

### DIFF
--- a/website/versioned_docs/version-5.x/ecosystem-angular.md
+++ b/website/versioned_docs/version-5.x/ecosystem-angular.md
@@ -4,6 +4,10 @@ title: single-spa-angular
 sidebar_label: Angular
 ---
 
+### Project Status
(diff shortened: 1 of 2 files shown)


filoxo (commented):
What do you think of also adding some sort of banner/callout/section that talks about how single-spa-angular is community-maintained and everything that comes with that (expected contributions/collaboration, potentially longer lead times on resolving issues, etc)? And maybe even still a call for contributors so that anyone who knows Angular internals well can also feel welcome to put in some effort.

internettrans:
Good idea, updated.

filoxo (approved):
🙊  Excellent!

PR #411, opened 2021-02-12
Fix: API uses location objects, not URL strings

timjb:


website/versioned_docs/version-5.x/api.md
```diff
@@ -396,20 +396,20 @@ const appsForACertainRoute = singleSpa.checkActivityFunctions({
 console.log(appsForACertainRoute); // ['app2']
 ```
 
-Will call every app's activity function with the `mockWindowLocation` and give you list of which applications should be mounted with that location.
+Will call every app's activity function with `location` and give you list of which applications should be mounted with that location.
```
internettrans: Since it's impossible to construct a `Location` object, I think we should clarify here that you can use a `URL` object here. The examples and documentation should be updated to reflect that, as that's a better pattern than providing the `pathname` object thing shown in the current examples:

```js
const appsForACertainRoute = singleSpa.checkActivityFunctions(new URL(location.href, '/app2'));
```

website/versioned_docs/version-5.x/api.md
```diff
@@ -396,20 +396,20 @@ const appsForACertainRoute = singleSpa.checkActivityFunctions({
...
 
 <h3>arguments</h3>
 
 <dl className="args-list">
-	<dt>mockWindowLocation: string</dt>
-	<dd>A string representing a window.location that will be used when calling every application's activity function to test if they return true.</dd>
+	<dt>location: Location</dt>
```
internettrans:
```suggestion
	<dt>url: URL</dt>
```

internettrans (approved):
The documentation for checkActivityFunctions is somewhat confusing, thanks for calling that out in this PR. I've left a couple comments about how we can make it even better by using `URL` objects instead of `Location` objects.

internettrans (changes requested):
Oops I didn't mean to approve the PR - we should make the changes in the above comments before merging.

website/versioned_docs/version-5.x/api.md
timjb: Good point. Done.

website/versioned_docs/version-5.x/api.md
timjb: Done.

internettrans (approved):
Thanks

PR #412, opened 2021-02-13
Update module-types.md

Tiberriver256:
As a person new to Microfrontend, it was awkward for me to have to shift mid-article from using the term Microfrontend to the acronym MFE. My personal opinion is that the acronym should be either introduced early in the article with an explanation (could be as simple as putting the acronym next to the term `Microfrontend (MFE)`) or not used in this article.


internettrans (approved):
Agreed, thanks

PR #413, opened 2021-02-15
docs(separating-applications): add monorepo example

petermikitsh:
I made a monorepo example: https://github.com/petermikitsh/learn-single-spa

If there's more you'd like to see in that repo, let me know. I'd be appreciative of any feedback to make it more "monorepo-y" to satisfy being an example.


internettrans (commented):
Thanks for the contribution - would you mind also adding it to the Community Examples section of https://github.com/single-spa/single-spa.js.org/blob/master/website/versioned_docs/version-5.x/examples.md?

I do not have time to look through your example repo, but am happy to merge this anyway.

filoxo (commented):
I was able to review your repo. The code looks good to me. If I were to ask for changes could you:
- add some documentation comparing your setup to [our Recommended Setup](https://single-spa.js.org/docs/recommended-setup/)
- add some documentation detailing if/how to achieve [independent deployments](https://single-spa.js.org/docs/faq/#can-single-spa-applications-be-deployed-independently)
    - https://github.com/single-spa/single-spa.js.org/pull/329 goes a little more in-depth into how independent deployments could be achieved but it seems to depend on Circle CI at the moment. I could very well be wrong though.

These are very soft requirements though. We could merge this even before you do those since those changes will be on your repo's side. 

I agree with Joel, this could also be linked to in the Examples page. That does make me wonder, what examples should be included in this page @joeldenning? Just any repo that is implemented with a monorepo tool?

PR #433, opened 2021-04-08
Update ecosystem-ember.md

stephencweiss:
Ember ecosystem now allows for npm/yarn and the `app.import` would then reference `node_modules`, not `bower_components`, if used.


internettrans (approved):
thanks

PR #435, opened 2021-04-09
Testing docs

### LINKED ISSUE #389: Document testing options for cross-microfrontend imports
This has come up enough times that I'm thinking it could be added to the documentation so that people can find it for themselves. As of right now, I haven't found anywhere in the docs that talks about how to handle microfrontend imports in a testing scenario. Here is one example of what the core team usually recommends: https://single-spa.slack.com/archives/C8R6U7MT7/p1597759987468600?thread_ts=1597759216.468300&cid=C8R6U7MT7

> To do unit tests with cross microfrontend imports, you have two options:
> 1. Publish the shared dependency to NPM. and install it into your node_modules so that it works in Jest.
> 2. Mock the shared dependency with jest’s moduleNameMapper

TheMcMurder:
I wanted to start some documentation on this to resolve #389 and help answer some common questions.

This is my first draft on this and I paired with Carlos on it.

I wanted to get something here to help and I felt that we needed some concrete examples of the exceptions to "normal" unit tests.

I also felt that when it comes to testing we don't have a great solution for integration tests which is a fairly significant piece of the story. If y'all have suggestions on a better way to document this please let me know.

### DIFF
--- a/website/versioned_docs/version-5.x/testing/e2e.md
+++ b/website/versioned_docs/version-5.x/testing/e2e.md
@@ -13,3 +13,24 @@ As microfrontends gain widespread adoption, testing tools will catch up and the 
 End to End (E2E) testing a single-spa [application](/docs/module-types#applications), [parcel](/docs/module-types#parcels), or [utility](/docs/module-types/#utilities) is very similar to E2E testing in other architectures. Because you are testing in the browser you can even use tools like [import-map-overrides](https://github.com/joeldenning/import-map-overrides) to run your tests in a production or production like environment with an override _before_ deloying to that environment. 
 
 In general we suggest only using E2E tests to test integration points between microfrontends and core functionality following principles of either the [testing pryamid](https://www.browserstack.com/guide/testing-pyramid-for-test-automation) or the [testing trophy](https://kentcdodds.com/blog/write-tests).
+
+## Testing Options
+
+In single spa there are more ways to test your code in a browser using tools like [cypress](https://www.cypress.io/). Two common approachs are to test individual applications by using 'standalone' mode and testing everything together, both provide value in different ways.
+
+### "E2E" testing with "standalone" mode
+
+While not perfect "standalone" mode offers a way to run individual single-spa applications, it can also be used to test a single-spa application. There are several limitations to take into account though, if the microfrontend relies upon configuration or initialization happening in your single-spa `root-config` you cannot test those areas in standalone mode without mocking. Standalone mode works by creating a custom single-spa root-config on the fly that will just render the one application, so the code is the same as if it were running in production but the configuration is different.
+
+### Testing everything together
+
+Much like E2E tests run in traditional SPA applications you can open a brower and run assertions using tools like [Cypress](https://www.cypress.io/). Taking this approach is mirroring a full end to end test. You are running the exact same code that is in the environment. With some configuration and tools like [import-map-overrides](https://github.com/joeldenning/import-map-overrides) you can set-up your testing environment to work with overrides to the import map and can run end-to-end tests before deploying your code changes to an environment.
+
+#### Configuring E2E tests to work with overrides
+
+At a high level you will need to do the following before your environment can utilize overrides in E2E tests
+
+1. Use a tool like [import-map-overrides](https://github.com/joeldenning/import-map-overrides)
+1. Get the built code on a publically accessible domain. Similar to a "review app"
+1. Configure your E2E testing environment to set the override
+1. Run the E2E tests
--- a/website/versioned_docs/version-5.x/testing/e2e.md
+++ b/website/versioned_docs/version-5.x/testing/e2e.md
@@ -1,0 +1,15 @@
+---
+id: e2e
+title: E2E testing
+sidebar_label: E2E testing
+---
+
+:::info
+
+As microfrontends gain widespread adoption, testing tools will catch up and the testing story will improve.
+
+:::
+
+End to End (E2E) testing a single-spa [application](/docs/module-types#applications), [parcel](/docs/module-types#parcels), or [utility](/docs/module-types/#utilities) is very similar to E2E testing in other architectures. Because you are testing in the browser you can even use tools like [import-map-overrides](https://github.com/joeldenning/import-map-overrides) to run your tests in a production or production like environment with an override _before_ deloying to that environment. 
+
+In general we suggest only using E2E tests to test integration points between microfrontends and core functionality following principles of either the [testing pryamid](https://www.browserstack.com/guide/testing-pyramid-for-test-automation) or the [testing trophy](https://kentcdodds.com/blog/write-tests).
--- a/website/versioned_docs/version-5.x/testing/units.md
+++ b/website/versioned_docs/version-5.x/testing/units.md
@@ -37,3 +37,73 @@ An example of this can be found in `people` and `planets` applications from `rea
 ## Shared Mocks
 
 If each project mocks every other microfrontend it is possible that the mocks will eventually become out of sync with the actual deployed microfrontend. One way to prevent this is to share mocks so that keeping the mocks in sync only requires one change instead of updating mocks in many different repositories.
+
+## Testing examples
+
+### Jest
+In the above examples I showed how `People` imports a component from `styleguide`. In order to unit test the component in people with Jest you will need to [configure](https://jestjs.io/docs/configuration) jest and mock the `styleguide` MFE. In jest configuration is done via multiple areas.
+1. [Create a jest config file](#jest-config)
+1. [Setup a mocks directory at the root](#mocks-directory)
+1. [Add a setupFile](#setup-file)
+
+#### Jest config
+Jest is configured with a configuration file. Jest supports many options of configuration, below is an example configuration using some of the options. See [more options on Jest's official documentation site](https://jestjs.io/docs/configuration).
+````js
+module.exports = {
+  collectCoverageFrom: ['src/**/*.js'],
+  modulePathIgnorePatterns: ['/cache', '/dist'],
+  transform: {
+    '^.+\\.(j|t)sx?$': 'babel-jest',
+  },
+  setupFilesAfterEnv: ['./jest.setup.js'],
+  moduleNameMapper: {
+    // Note this is only needed if you don't match the module name directly
+    // an alternative would be to place the mock in 
+    // <rootDir>/__mocks__/@react-mf/styleguide.js and it would be autodetected
+    '@react-mf/styleguide': '<rootDir>/__mocks__/styleguide.js',
+  },
+}
+````
+
+#### mocks directory
+Jest will detect folders named `__mocks__` and if the naming convention is exact or the modules have been mapped using `moduleNameMapper` then Jest will use those mocks in place of an import. This structure is essential for other microfrontends where you don't have the code locally. [See more information on jest's official documentation](https://jestjs.io/docs/manual-mocks)
+```
+.
+├── __mocks__
+│   └── styleguide.js
+├── src
+│   ├── react-mf-people.js
+│   └── ...
+├── node_modules
+├── jest.setup.js
+├── ...
+└── jest.config.json
+```
+
+#### setup file
+Jest uses a setup file to create globals mocks that can be utilized by every test or otherwise configure the test environment. If you were mocking `localStorage` or `SystemJS` this is a good place to configure those mocks. [See more use-cases for a set-up file on Jest's offical documentation](https://jestjs.io/docs/configuration#setupfilesafterenv-array)
+```js
+// jest.setup.js
+// import Mocks for SystemJS mock below
+import peopleApplication from '@react-mf/people'
+// Mock SystemJS
+global.System = {
+  import: jest.fn(mockImport)
+}
+
+function mockImport (importName) {
+  // What you do in mock import will depend a lot on how you use SystemJS in the project and components you wish to test
+
+  /* If I had already created a mock for `@react-mf/people` and I wanted to test this component:
+  *  https://github.com/react-microfrontends/planets/blob/main/src/planets-page/selected-planet/selected-planet.component.js#L5
+  * I would want `System.import('@react-mf/people')` to resovle to my mock one way to accomplish this would be the following
+  */
+  if (importName === '@react-mf/people') {
+    return Promise.resolve(peopleApplication)
+  } else {
+    console.warn('No mock module found')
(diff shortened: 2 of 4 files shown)


website/versioned_docs/version-5.x/testing/e2e.md
```diff
@@ -0,0 +1,15 @@
...
+sidebar_label: E2E testing
+---
+
+:::info
+
+As microfrontends gain widespread adoption, testing tools will catch up and the testing story will improve.
+
+:::
+
+End to End (E2E) testing a single-spa [application](/docs/module-types#applications), [parcel](/docs/module-types#parcels), or [utility](/docs/module-types/#utilities) is very similar to E2E testing in other architectures. Because you are testing in the browser you can even use tools like [import-map-overrides](https://github.com/joeldenning/import-map-overrides) to run your tests in a production or production like environment with an override _before_ deloying to that environment.
```
internettrans, line 13: Let's address whether e2e testing should be done on one microfrontend at a time or not. That's the main question people have. If it's done for one microfrontend at a time, the question is how to get that set up so that it still works even with cross microfrontend imports. The standalone plugin can help in that scenario.

Also, cypress.io is the most popular e2e testing framework right now I think, so we could maybe link to it.

website/versioned_docs/version-5.x/testing/units.md
```diff
@@ -0,0 +1,39 @@
...
+sidebar_label: Unit testing
+---
+
+:::info
+
+As microfrontends gain widespread adoption, testing tools will catch up and the testing story will improve.
+
+:::
+
+Unit testing a single-spa [application](/docs/module-types#applications), [parcel](/docs/module-types#parcels), or [utility](/docs/module-types/#utilities) is very similar to unit testing in the framework you are using with two notable exceptions:
```
internettrans:
```suggestion
Unit testing a single-spa [application](/docs/module-types#applications), [parcel](/docs/module-types#parcels), or [utility](/docs/module-types/#utilities) is very similar to unit testing in the framework you are using, with two notable exceptions:
```

website/versioned_docs/version-5.x/testing/units.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Importing microfrontends
+
+It is fairly common in microfrontends to have one microfrontend import and rely upon a component from another microfrontend. Reliance on another microfrontend can be challenging to test because unit tests generally run locally and you won't have access to other microfrontends. When this occurs we generally recommend mocking the other microfrontend for the unit test.
+
+An example of this can be found in the `Button` component exported by the `styleguide` in [react.microfrontends.app](https://github.com/react-microfrontends/styleguide/blob/master/src/button.component.js). Because that component is imported and used by the [planets application](https://github.com/react-microfrontends/planets/blob/41ba0aaf9005b5300cc28ad5f4eac024eae06e2b/src/planets-page/planets-page.component.js#L6) you will need to make it available to the test environment by mocking the dependency. This is necessary because the test environment cannot dynamically import other microfrontends, like the browser can. Given the wide variety of unit testing tools you will need to follow the pattern established by the test environment you are using for mocking other microfrontends.
```
internettrans, line 23: Sharing a code example of how to mock a file in jest will be very helpful for a lot of people. People don't know about moduleNameMapper,  nor do they have the patterns in place for a `__mocks__` directory with `__mocks__/styleguide.js` mock file

website/versioned_docs/version-5.x/testing/units.md
```diff
@@ -0,0 +1,39 @@
...
+
+:::
+
+## `System.import`
+
+Occasionally you will choose to interop with another microfrontend asynchronously by explicitly calling `System.import`. Testing in this scenairo may require mocking both SystemJS and the module you're importing. Additionally because `System.import` returns a promise your tests in that area will need to be asynchronous and wait for promises to resolve.
+
+An example of this can be found in `people` and `planets` applications from `react.microfrontends.app`. The [People application](https://github.com/react-microfrontends/people/blob/master/src/react-mf-people.js#L21) exports a function that resolves with a component. The [Planets Application](https://github.com/react-microfrontends/planets/blob/main/src/planets-page/selected-planet/selected-planet.component.js) imports and uses that component asynchronously with `React.lazy`. Testing this component would necessitate mocking both `SystemJS` and `People`.
```
internettrans, line 35: Explaining how to mock SystemJS would be helpful. Including how to configure jest to run a setup-tests.js file.

```js
// setup-tests.js

global.System = {
  "import": () => ({
      Button: (props) => <button>{...props}</button>
   })
 }
```

internettrans (commented):
This will be very helpful documentation, a couple small comments

website/versioned_docs/version-5.x/testing/e2e.md
TheMcMurder: I actually struggled with this part which is why I "skipped it". I'm torn on how to suggest it to work.

The part that's getting me confused and twisted is how to conceptualize them.

Unit tests - Jest
Integration tests - ?
End to End tests - Cypress in a real environment

I think it's arguable that you could also do integration tests with Cypress in a real environment to test the integration between modules but I don't know if that's the best way to do it.

What would be _killer_ is a JSDOM environment that SystemJS could run in and you could run Integration tests without running a browser, but that gets complicated as well.

internettrans: I think our aim should be just to explain options, rather than prescribing which one fits best for everyone. We could talk about using import-map-overrides query params for e2e tests on a deployed environment, and also talk about using the standalone plugin + cypress as a way of testing things one at a time.

website/versioned_docs/version-5.x/testing/units.md
```diff
@@ -0,0 +1,109 @@
...
+.
+├── __mocks__
+│   └── styleguide.js
+├── src
+│   ├── react-mf-people.js
+│   └── ...
+├── node_modules
+├── jest.setup.js
+├── ...
+└── jest.config.json
```
internettrans: I think this should be `jest.config.js`, since the example file above is a CJS file, not json.

website/versioned_docs/version-5.x/testing/e2e.md
TheMcMurder: This relates to why I struggled to expound on this area.

In my mind the definition of End to End tests is roughly:

> Tests that simulate what user sees and does. They should test everything in a production-like environment making actual network requests, using an actual browser, etc.

So I struggle with using `standalone` mode for this because it's not a production-like environment. I see standalone mode as a potential way to do integration tests but I worry that it's too limited to actually test many different integration points.

Perhaps my approach to this is too strict/limited and I'm not seeing the value of standalone mode.

TheMcMurder: So in a vacuum I'd say something like this:

> Standalone mode is not recommended for writing E2E tests. We suggest all E2E tests run in an environment that is a close to production as possible and while standalone mode has other values it cannot deliver an experience that is close enough to production to effectively test in an end to end scenario.

But I'm not sure that's right and I wonder if the definitions I've been using to help me understand testing are wrong or don't translate well to MFEs

internettrans: I don't think standalone mode is discouraged for e2e tests - it's an option that some are already doing and have found useful.

TheMcMurder: Yup, that's my point. I have a bit of a block on standalone mode. I don't see how it's actually an end to end test in standalone mode. End to end to me implies no mocking, but standalone mode only works _because of mocking_

internettrans: Standalone mode doesn't mock anything - it generally uses a live, deployed import map to resolve cross microfrontend imports. The main difference between standalone mode and a deployed environment is in the application registrations - on a deployed environment it registers all the applications whereas in standalone mode it only registers the one application. The benefit of standalone mode is basically just "get a root config for free that lets you see just this one microfrontend." But it doesn't require any mocking.

TheMcMurder:
@joeldenning I did a first pass on adding options related to E2E tests. This should be ready for another review.

website/versioned_docs/version-5.x/testing/e2e.md
```diff
@@ -0,0 +1,36 @@
...
```
internettrans, line 13:
```suggestion
End to End (E2E) testing a single-spa [application](/docs/module-types#applications), [parcel](/docs/module-types#parcels), or [utility](/docs/module-types/#utilities) is very similar to E2E testing in other architectures. Because you are testing in the browser you can even use tools like [import-map-overrides](https://github.com/joeldenning/import-map-overrides) to run your tests in a production or production like environment with an override _before_ deploying to that environment. 
```

website/versioned_docs/version-5.x/testing/e2e.md
```diff
@@ -0,0 +1,36 @@
...
+
+:::
+
+End to End (E2E) testing a single-spa [application](/docs/module-types#applications), [parcel](/docs/module-types#parcels), or [utility](/docs/module-types/#utilities) is very similar to E2E testing in other architectures. Because you are testing in the browser you can even use tools like [import-map-overrides](https://github.com/joeldenning/import-map-overrides) to run your tests in a production or production like environment with an override _before_ deloying to that environment. 
+
+In general we suggest only using E2E tests to test integration points between microfrontends and core functionality following principles of either the [testing pryamid](https://www.browserstack.com/guide/testing-pyramid-for-test-automation) or the [testing trophy](https://kentcdodds.com/blog/write-tests).
+
+## Testing Options
+
+In single spa there are more ways to test your code in a browser using tools like [cypress](https://www.cypress.io/). Two common approachs are to test individual applications by using 'standalone' mode and testing everything together, both provide value in different ways.
```
internettrans, line 19:
```suggestion
In single-spa, there are more ways to test your code in a browser using tools like [cypress](https://www.cypress.io/). Two common approaches are to test individual applications by using standalone mode and testing everything together, both provide value in different ways.
```

website/versioned_docs/version-5.x/testing/e2e.md
```diff
@@ -0,0 +1,36 @@
...
+
+### "E2E" testing with "standalone" mode
+
+While not perfect "standalone" mode offers a way to run individual single-spa applications, it can also be used to test a single-spa application. There are several limitations to take into account though, if the microfrontend relies upon configuration or initialization happening in your single-spa `root-config` you cannot test those areas in standalone mode without mocking. Standalone mode works by creating a custom single-spa root-config on the fly that will just render the one application, so the code is the same as if it were running in production but the configuration is different.
```
internettrans, line 23: The quotes around standalone seem unnecessary, and you only listed one limitation so I suggested removing the part that says there are several limitations.

```suggestion
While not perfect, standalone mode offers a way to run individual single-spa applications and can be used to test a single-spa application. If the microfrontend relies upon configuration or initialization happening in your single-spa `root-config` you cannot test those areas in standalone mode without mocking. Standalone mode works by creating a custom single-spa root-config on the fly that will just render the one application, so the code is the same as if it were running in production but the configuration is different.
```

TheMcMurder:
I'm going to merge this as "good enough" and we can iterate on it.

website/versioned_docs/version-5.x/testing/units.md
```diff
@@ -0,0 +1,109 @@
...
+## Testing examples
+
+### Jest
+In the above examples I showed how `People` imports a component from `styleguide`. In order to unit test the component in people with Jest you will need to [configure](https://jestjs.io/docs/configuration) jest and mock the `styleguide` MFE. In jest configuration is done via multiple areas.
+1. [Create a jest config file](#jest-config)
+1. [Setup a mocks directory at the root](#mocks-directory)
+1. [Add a setupFile](#setup-file)
+
+#### Jest config
+Jest is configured with a configuration file. Jest supports many options of configuration, below is an example configuration using some of the options. See [more options on Jest's official documentation site](https://jestjs.io/docs/configuration).
```
internettrans, line 50:
```suggestion
Jest is configured with a configuration file. Below is an example configuration using some of the options. See [more options on Jest's official documentation site](https://jestjs.io/docs/configuration).
```

website/versioned_docs/version-5.x/testing/units.md
```diff
@@ -0,0 +1,109 @@
...
+    // Note this is only needed if you don't match the module name directly
+    // an alternative would be to place the mock in 
+    // <rootDir>/__mocks__/@react-mf/styleguide.js and it would be autodetected
+    '@react-mf/styleguide': '<rootDir>/__mocks__/styleguide.js',
+  },
+}
+````
+
+#### mocks directory
+Jest will detect folders named `__mocks__` and if the naming convention is exact or the modules have been mapped using `moduleNameMapper` then Jest will use those mocks in place of an import. This structure is essential for other microfrontends where you don't have the code locally. [See more information on jest's official documentation](https://jestjs.io/docs/manual-mocks)
```
internettrans, line 69: This isn't true - jest doesn't automatically look for `__mocks__`. You have to specify the folder in the moduleNameMapper

internettrans (commented):
Looks good - last round of comments

website/versioned_docs/version-5.x/testing/units.md
TheMcMurder: https://github.com/react-microfrontends/planets/pull/20

I don't have `moduleNameMapper` specified [here](https://github.com/react-microfrontends/planets/pull/20/files#diff-1e058ca1442e46581b13571fb8d261f9e1f5657e26c96634d4c1072f0f0347f1R1) at all.

TheMcMurder: I'm guessing this is the result of poor communication on my part. 

Would something like the following be more clear?

## Mocks Directory

Mocking imports of microfrontends in jest can be done in three ways. 
1. Use `moduleNameMapper` to map mock files to their import name. 
````
// jest.config.js
...8
moduleNameMapper: {
    '@react-mf/api': '<rootDir>/__mocks__/api.js',
  },
````
2. Use jest folder structure convention.
 - If you create a directory at the root of the project and name it `__mocks__` you can then mock your modules by name in that directory. For example if I wanted to mock `@react-mf/api` which has a `/`you would have a directory structure like this:
```
- __mocks__
--- @react-mf
------ api.js
```
3. Mock in jest setup files.

internettrans: Ah I didn't know that, and stand corrected - that's a cool feature of jest.

PR #464, opened 2021-06-29
chore: add deutsche telekom logo to Users

Lalaluka:
Hi together ✌️,

after some rounds with @joeldenning we finally figured out how to find a good way to add the Deutsche Telekom Logo to the single-spa website without running into Licensing Problems.

If any issues occur the internal Approval Number of the Brand Project is: `P25926`
(Joel also holds a license to use the logo for this purpose)

For now, I didn't add us to the pinned companies. But we would be delighted to be pinned of course 😉


internettrans:
Let's make this one pinned - looks good!

Lalaluka:
Just added the pinned 👍

internettrans (approved):
Thanks!

PR #466, opened 2021-07-11
Added a commit to add the example

nitinreddy3:
Examples that covers the route based rendering of the application -> react and angular


internettrans (approved):
Thanks

PR #470, opened 2021-07-29
feat: improve accessibility of links

DaniAcu:
It's hard to find the link in the documentation. So I improve the styles to do them more accessible for everyone.

Before
image

After
image


internettrans (approved):
Thanks, yes this has been a problem since the branding update

PR #480, opened 2021-08-13
fix: flavored markdown creating email links

Lalaluka:
HI,
On the ecosystem-vue Page, I encountered two occasions of `single-spa-vue@X.X.X` being rendered as `mailto` Links:
Bildschirmfoto 2021-08-13 um 10 30 15
Bildschirmfoto 2021-08-13 um 10 31 23

This is probably because of the markdown flavor. I could fix that by adding a span until after the @ 
`<span>single-spa-vue@</span>2.4.0`
Which disables the autogeneration of the link:
Bildschirmfoto 2021-08-13 um 10 33 08

It's a pretty minor fix but also helps accessibility because such links are getting targeted by navigating the side through the keyboard or Screenreaders.

If such issues are also present on other pages the same fix would work too. If this improvement is wanted I can take a closer look.


internettrans (approved):
Thanks

PR #486, opened 2021-09-11
add enabling hash mode in routing to ecosystem-angular.md

zenan-xu:
Hi,
After this discussion https://single-spa.slack.com/archives/C8R6U7MT7/p1631087971271300 in slack, I would like to complete the routing part of ecosystem angular with `Enabling hash mode` .
Here is a simple example showing how to do this with angular https://github.com/zenan-xu/single-spa-hash-mode-example-with-angular.
Thanks


internettrans (approved):
Thanks

PR #491, opened 2021-09-30
add additional examples to angular parcel

jsmapr1:
Hello, I added some additional examples to the angular parcel documentation.

1) Importing an application that is defined on an import map.

Screen Shot 2021-09-30 at 9 55 40 AM

2) Passing custom props to the parcel.

Screen Shot 2021-09-30 at 9 53 53 AM

Let me know what you think and if you need any additional details.

### DIFF
--- a/website/versioned_docs/version-5.x/ecosystem-angular.md
+++ b/website/versioned_docs/version-5.x/ecosystem-angular.md
@@ -939,7 +939,7 @@ export const config = singleSpaReact({
 ### Loading External Components Asynchronously
 
 If your code is part of an import map and you include a global reference to
-systemjs, you can dynamically set the as an async function. 
+systemjs, you can dynamically import the code using an `async` method.
 
 For example, if your import map includes the `ReactComponent`.
 
@@ -952,7 +952,7 @@ For example, if your import map includes the `ReactComponent`.
 ```
 
 You can dynamically load the component by setting the `config` as an
-asynchronous function that fetches the component.
+asynchronous method that fetches the component.
 
 Since some versions of webpack use SystemJS under the hood, you'll need to
 reference the global version.
--- a/website/versioned_docs/version-5.x/ecosystem-angular.md
+++ b/website/versioned_docs/version-5.x/ecosystem-angular.md
@@ -936,6 +936,101 @@ export const config = singleSpaReact({
 });
 ```
 
+### Loading External Components Asynchronously
+
+If your code is part of an import map and you include a global reference to
+systemjs, you can dynamically set the as an async function. 
+
+For example, if your import map includes the `ReactComponent`.
+
+```javascript
+{
+  imports: {
+    "@org/react-component": '/org-react-component.js'
+  }
+}
+```
+
+You can dynamically load the component by setting the `config` as an
+asynchronous function that fetches the component.
+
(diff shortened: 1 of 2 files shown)


website/versioned_docs/version-5.x/ecosystem-angular.md
```diff
@@ -936,6 +936,101 @@ export const config = singleSpaReact({
...
+  mountRootParcel = mountRootParcel;
+}
+```
+
+### Passing Custom Props
+
+You can pass any custom props to the parcel by passing an object of props using
+the `customProps` attribute.
+
+If you have a custom click event in your React component:
```
internettrans: It's a bit unclear why the angular documentation is showing a react parcel. That is a common way to use it, but perhaps even more common is to have angular parcels rendered by angular applications (since sharing Angular versions is very difficult and upgrading between versions is very difficult). Perhaps just clarify that you're talking about an example with a React parcel:

```suggestion
For example, if you're rendering a React parcel from an Angular component, you can pass a click handler from Angular into the React parcel:
```

website/versioned_docs/version-5.x/ecosystem-angular.md
```diff
@@ -936,6 +936,101 @@ export const config = singleSpaReact({
...
+const ReactWidget = ({ handleClick }) => <button onClick={handleClick}>Click Me!</button>;
+
+export const config = singleSpaReact({
+  React,
+  ReactDOM,
+  rootComponent: ReactWidget,
+});
+```
+
+You can pass a function as a custom prop. Remember that the context changes when
```
internettrans: > Remember that the context changes when you pass in a function

This is not technically correct, although I agree with the spirit of trying to help people avoid common pitfalls. So maybe let's reword it to be correct. Passing a function around does not change its context - in javascript the context is caller-dependent by default, meaning that whoever invokes the function gets to determine what the context is. The exception is for bound functions, whose context is set by the creator of the function and cannot be altered by the caller. The reason that `handleClick = () => {` is better is because that syntax creates a function that is bound to the current instance of the class, not because it's an arrow function (arrow functions follow the same rules and have caller-dependent context, by default)

Maybe something like this:

```suggestion
You can pass a function (or any other value) as a custom prop. To ensure that the functions you pass to the parcel are bound with the correct javascript context, use the `handleClick = () => {` syntax when defining your functions.
```

website/versioned_docs/version-5.x/ecosystem-angular.md
```diff
@@ -936,6 +936,101 @@ export const config = singleSpaReact({
...
+
+```tsx
+// Inside of src/app/ReactWidget/ReactWidget.tsx
+import * as React from 'react';
+import * as ReactDOM from 'react-dom';
+import singleSpaReact from 'single-spa-react';
+
+const ReactWidget = ({ handleClick }) => <button onClick={handleClick}>Click Me!</button>;
+
+export const config = singleSpaReact({
```
internettrans:
```suggestion
export const parcelConfig = singleSpaReact({
```

internettrans (commented):
Thanks for your help with this, I've left a few minor comments.

jsmapr1:
updated

internettrans (approved):
Thanks

PR #493, opened 2021-10-11
Remove config duplication in create-single-spa.md

MartynRobotham:
Remove duplicated config in webpack-config-single-spa-react-ts section of create-single-spa.md & add in missing comma into same section


filoxo (approved):
Thank you for fixing these mistakes!

PR #498, opened 2021-11-04
Another option for the CI/CD 

Hamels:
In our setup we make use of http redirect 302. So we don't have to update the importmap. It's an idea which I didn't see in the documentation.

### DIFF
--- a/website/versioned_docs/version-5.x/recommended-setup.md
+++ b/website/versioned_docs/version-5.x/recommended-setup.md
@@ -288,7 +288,7 @@ The advantage of a) is that it is concurrent-safe for multiple, simultaneous dep
 
 The advantage of b) is that it doesn't require running the import-map-deployer in your production environment. Ultimately, you should choose whichever option makes sense for your organization.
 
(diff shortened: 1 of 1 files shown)


website/versioned_docs/version-5.x/recommended-setup.md
```diff
@@ -288,6 +288,8 @@ The advantage of a) is that it is concurrent-safe for multiple, simultaneous dep
 
 The advantage of b) is that it doesn't require running the import-map-deployer in your production environment. Ultimately, you should choose whichever option makes sense for your organization.
 
+Another option of deploying and making sure the latests javascript files are beeing used is making use of redirect (http statuscode 302). The single-spa importmap uses the unhashed url of the javascript file. eg: dist/app.js. Then when the requests for this file comes to the server it is redirected to the actual deployed file eg. dist/app.123abc.js and this is then served to the client. This way the importmap never has to be updated and the microfronted can be separatly deployed.
```
internettrans:
```suggestion
Another option of deploying and making sure the latests javascript files are beeing used is making use of redirect (HTTP status code 302). The single-spa import map uses the unhashed url of the javascript file. eg: dist/app.js. Then when the requests for this file comes to the server it is redirected to the actual deployed file eg. dist/app.123abc.js and this is then served to the client. This way the import map never has to be updated and the microfrontend can be separately deployed.
```

internettrans (approved):
Thanks, agreed that this is a valid approach. The redirect is a slight performance hit (extra network request) but solves a lot of problems and is probably worth it to solve those.

PR #506, opened 2021-12-23
[angular-ecosystem] add how to correct angular i18n error in polyfills part

zenan-xu:
Hi,
I recently faced the follwing error. 
Error_i18n
For my case, it was caused by one of my MF using datepicker and paginator of ng-bootstrap which use i18n.
Normaly, we can import `@angular/localize/init` to the polyfills of angular application to correct this error. 

As polyfills in angular MF application will not be loaded automatically, the way I found is to install `@angular/localize` to root-config module and import `@angular/localize/init` to the root-config.js.

So I would like to complete polyfillls part with this correction if anyone else will face the same error.
Thanks

### DIFF
--- a/website/versioned_docs/version-5.x/ecosystem-angular.md
+++ b/website/versioned_docs/version-5.x/ecosystem-angular.md
@@ -773,6 +773,17 @@ If you're looking for a quick one-liner, try adding this line near the top of yo
 <script src="https://unpkg.com/core-js-bundle/minified.js"></script>
 ```
 
+To correct the error `It looks like your application or one of its dependencies is using i18n`.
+
+Install `@angular/localize` in your root-config module
+```sh
+npm i @angular/localize
+```
+Add following import to your root-config.js
+```ts
+import "@angular/localize/init";
+```
+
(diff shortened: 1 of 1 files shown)


website/versioned_docs/version-5.x/ecosystem-angular.md
```diff
@@ -773,6 +773,17 @@ If you're looking for a quick one-liner, try adding this line near the top of yo
...
 
+To correct the error `It looks like your application or one of its dependencies is using i18n`.
+
+Install `@angular/localize` in your root-config module
+```sh
+npm i @angular/localize
+```
+Add following import to your root-config.js
+```ts
+import "@angular/localize/init";
```
internettrans, line 784: Hmm it might be best to add a note that `@angular/localize` (and other angular libs) should be shared dependencies, to avoid redownloading them in root config and in other MFEs. However, that can be done in a follow-up PR. Please send that follow up PR if interested in contributing further.

internettrans (approved):
Thank you for contribution

PR #508, opened 2021-12-30
Added Kvist to "Who's Using This?"

MathiasWP:
I am the CTO of a Norwegian-based startup called Kvist Solutions, and I speak for our whole company when I say: **Thank you single-spa!**

All the things I've learned from using single-spa and reading your documentation has resulted in an elegant frontend architecture for us. We rely heavily on some of the core principles listed in the single-spa documentation, and it was an obvious choice to use this library as the underlying engine for the system we've developed for our enterprise application.

We are proud users of this library, and would gladly be a part of the "users"-list.


internettrans (approved):
Thanks @MathiasWP for the kind words, and welcome!

website/src/data/users.js
```diff
@@ -125,6 +125,12 @@ const users = [
     infoLink: 'https://devapi.com.br',
     pinned: false
   },
+    {
```
internettrans:
```suggestion
  {
```

Issue #507, opened 2021-12-26
Browser Search marks minimized menus as find


Lalaluka:
Using the Browser STRG+F search in the docs, it returns minimized/hidden menus as search results (even tho they are not visible).
I'm not sure, but it might be an issue with the used docusaurus version?

Example Results from _"Getting Started with single-spa"_
<img width="357" alt="Bildschirmfoto 2021-12-26 um 21 30 46" src="https://user-images.githubusercontent.com/26364692/147419345-80bf2e4e-2920-4790-b2f5-1d8c6fadaede.png">
<img width="357" alt="Bildschirmfoto 2021-12-26 um 21 36 42" src="https://user-images.githubusercontent.com/26364692/147419464-328226a6-5553-4d49-9fe1-13287056426c.png">

Menu State during the search:
<img width="288" alt="Bildschirmfoto 2021-12-26 um 21 41 20" src="https://user-images.githubusercontent.com/26364692/147419551-15dab6d5-2002-4388-9972-f0cdd05996a6.png">

After extending, the search result counter (1/1) won't increase. But It will mark the menu occurrence.
<img width="288" alt="Bildschirmfoto 2021-12-26 um 21 43 23" src="https://user-images.githubusercontent.com/26364692/147419594-4f1c5b51-289e-4dbf-bc2f-df05e9b36366.png">
<img width="288" alt="Bildschirmfoto 2021-12-26 um 21 45 04" src="https://user-images.githubusercontent.com/26364692/147419627-5b1fde4e-d9c7-432e-a64c-1338fc0e70f0.png">

Reproducible on:

- Chrome 96.0.4664.55
- Opera 74.0.3911.218
- Edge 96.0.1054.62

Safari seems to work fine.

internettrans:
Agreed that this could be improved. I don't know the specifics of what browsers will find when searching via Ctrl / Cmd F. Is it just everything in the DOM? Or only things with certain css properties? Regardless, this should be raised as an issue in Docusaurus rather than here. I wasn't able to find any existing issue in docusaurus' queue for this, so perhaps you could create a new one?

https://github.com/facebook/docusaurus/issues

PR #509, opened 2022-01-05
chore: Upgrade Docusaurus Version

Lalaluka:
This seems like a redundant change. 
But before adding an issue Ticket to Docusaurus I just upgraded the Version of Docusaurus from `2.0.0-alpha.70` to `^2.0.0-beta.14`.

This fixes the Bug described in #507 

I won't file another Issue in Docusaurus since this already takes care of.
The diff is of course quite large since it updates almost the full lock file.

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -19,8 +19,8 @@
     "prettier": "^1.19.1"
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


internettrans (approved):
This is probably a big change. Here's to hoping that it doesn't cause major breakages :)

PR #510, opened 2022-01-08
Fix broken link

Lalaluka:
The first Link in the Warning on https://single-spa.js.org/docs/ecosystem-vue/#vue-3 is broken and leads to the 404 Page.
This PR fixes the link.

Tested locally and works now.


filoxo (approved):

filoxo:
Thanks for fixing this!

PR #527, opened 2022-02-08
fix: typo in building-applications

Lalaluka:
In [buildingApplications](https://single-spa.js.org/docs/building-applications) the heading `Lifecyle props` probably has a Typo ;) 
This PR fixes all links (some already had the right version and were broken in the previous version) and the heading itself.

This also requires a Change in [vue-cli-plugin-single-spa](https://github.com/single-spa/vue-cli-plugin-single-spa) since the plugin generates a comment referencing this page. I will create a PR with the change in the repository, but the changes should probably be merged at the same time.

Related issue single-spa/vue-cli-plugin-single-spa#42


frehner (approved):
Thanks!

PR #541, opened 2022-04-29
fix activity function link in building applications

cmgchess:
Currently the `Activity Function` link in the `Building Applications` page leads to a broken page
https://single-spa.js.org/docs/building-applications/
image


Lalaluka (approved):

Lalaluka:
Thank you for the fix!

PR #542, opened 2022-04-29
fix how to contribute link in devtools

cmgchess:
Added the https://single-spa.js.org/docs/contributing-overview/#how-to-contribute link to https://single-spa.js.org/docs/devtools


Lalaluka (approved):
LGTM. Thanks for the fix!

PR #543, opened 2022-04-29
fix mailto link

cmgchess:
A mailto link has been created because of the @ in https://single-spa.js.org/docs/devtools


Lalaluka (approved):
LGTM! Thanks.

Just a short heads up. It's easier and less work for maintainers if small typo fixes like these are collected in a single PR instead of a split in very small PRs.

But thank you for your contributions!

cmgchess:
sure I'll keep that in mind

PR #546, opened 2022-05-07
Add `vite-ts-single-spa-root-config` example.

long-woo:


Lalaluka (approved):
LGTM!
Thanks for this example ❤️

PR #575, opened 2023-01-18
Update Docusaurus

filoxo:
in my last PR, i noticed that the docs site was still using an alpha version of docusaurus@2 from 3 years ago. docusarurus@2 final was released late last year (https://docusaurus.io/blog/2022/08/01/announcing-docusaurus-2.0) so the latest should be stable enough to adopt.


.nvmrc
```diff
@@ -0,0 +1 @@
+>=16.14
```
filoxo, line 1: This will only be use if a user has nvm. Otherwise, the user's package manager will warn them if they're using an incorrect version, with a message similar to:

```sh
error @docusaurus/core@2.2.0: The engine "node" is incompatible with this module. Expected version ">=16.14". Got "16.13.0"
```

filoxo:
ON HOLD: i failed to push a commit that actually updated the lockfile. at this point i discovered that my update wasn't complete and Algolia requires an appId now.

PR #600, opened 2023-10-23
JP/single-spa-react

webJose:
Update example for React versions lower than v18.


webJose:
Here are some screenshots of the changes.  Apologies, I was test driving the GitHub PR extension in VS Code.  It works nicely, but I could not add these shots in the original message.

image

image

internettrans:
I think we should fix latest single-spa-react to work with React 17, rather than update the documentation for it. See https://github.com/single-spa/single-spa-react/issues/202

internettrans:
This documentation change is no longer necessary, due to the fix in single-spa-react@6.0.1

https://github.com/single-spa/single-spa-react/releases/tag/v6.0.1

PR #606, opened 2023-11-18
Fix HTTP request docs for SSR

PepijnSenders:
`sendLayoutHTTPResponse` only takes one argument, and that's the `RenderOptions` object.


MilanKovacic:
Hi, thank you for the PR. Could you please adapt it since v6 of the documentation has been released?

PR #626, opened 2024-03-10
docs: add angularjs example

mbarbosasan:
Hi guys, first of all, sorry if i make some grammar mistakes, english isn't my primary language.

In the past year, i had to update a legacy application with a codebase 10 years old that was using AngularJS, and unfortunately, we can't just throw away all components and rewrite them in new frameworks, we found SingleSPA and Microfrontends and was perfect for what we are searching, but we had some struggle to find content about SingleSPA with AngularJS.

So i would like to give back and have created a simple starter for those who will or are trying to do so, maybe it's not the best configs but certainly will help somebody.


MilanKovacic (approved):

MilanKovacic:
Thank you for the contribution!

PR #628, opened 2024-03-14
Fixed typo in parameter `spaEntryPoints` in ecosystem-vite.md

ashylen:
As per `vite-plugin-single-spa` docs: https://github.com/WJSoftware/vite-plugin-single-spa/blob/4a96185b15c61e0aa9399568b793f0e6c687bce6/README.md?plain=1#L248 parameter here should be `spaEntryPoints` (plural) instead of `spaEntryPoint`.


webJose:
Hello.  It is not a typo.  It merely was how it was in earlier versions of the plug-in.  I just submitted PR #639 to update the docs to v0.7.0.  You should close this PR.

PR #634, opened 2024-04-09
docs: upgrade Angular guide

arturovt:
I've added a warning about the importance of updating the builder before running schematics.

### DIFF
--- a/versioned_docs/version-6.x/ecosystem-angular.md
+++ b/versioned_docs/version-6.x/ecosystem-angular.md
@@ -36,7 +36,7 @@ For instructions on how to test this locally before creating a pull request, see
 
 ## Angular versions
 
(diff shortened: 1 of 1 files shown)


MilanKovacic:
Can we still retain the documentation for older Angular versions, but make them less visible?
A lot of consumers that are on very old Angular versions use single-spa to gradually migrate/rewrite applications to either new versions of Angular, or other frameworks.

internettrans:
I agree that keeping the documentation for older versions of Angular is best - some projects are stuck on old versions of Angular and hope single-spa will help them convert the monolith into a microfrontend so they can start new microfrontends with an updated framework.

arturovt:
@MilanKovacic can you have a look again

internettrans (approved):
Looks good to me

PR #636, opened 2024-05-03
(docs) Add example: kfrederix/single-spa-angular-esm

kfrederix:
Hello,

First of all: thank you for creating and maintaining single-spa! We have been using it for the past 4 years in the company I work for, and I am a big fan!

I have now published an example Nx monorepo, demonstrating how single-spa can be used for microfrontends with the new Angular builder based on Vite/esbuild. It leverages native browser importmaps, ES modules and makes use of [import-map-injector](https://github.com/single-spa/import-map-injector) to add support for having multiple/external importmaps.

I have also written [extensive documentation](https://github.com/kfrederix/single-spa-angular-esm/blob/main/doc/background.md) about specific issues/problems I faced while working on this example.

I would love to add this to the Examples section on the single-spa docs, if you think it could be helpful to others.

### DIFF
--- a/versioned_docs/version-6.x/examples.md
+++ b/versioned_docs/version-6.x/examples.md
@@ -26,6 +26,7 @@ sidebar_label: Resources
 
 ## Community examples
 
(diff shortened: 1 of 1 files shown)


MilanKovacic (approved):

MilanKovacic:
Thank you for the PR. Example looks valuable!

PR #648, opened 2024-09-21
docs: add unmounting styles in Angular

arturovt:


internettrans (approved):

versioned_docs/version-6.x/ecosystem-angular.md
```diff
@@ -790,6 +790,81 @@ be loaded by single-spa-angular's webpack config, without you having to configur
...
+
+```ts
+// @ts-ignore
+import unmountableStyles from './main.scss?unmountable';
+
+const lifecycles = singleSpaAngular({
+  bootstrapFunction: async () => {
+    unmountableStyles.use();
+    const appRef = await bootstrapApplication(AppComponent, appConfig);
+    appRef.onDestroy(() => unmountableStyles.unuse());
```
internettrans, line 858: Why don't the `unmountableStyles.use()` and `unmountableStyles.unuse()` calls happen inside of the single-spa-angular, rather than the bootstrapFunction?

I'd expect something like this:

```ts
const lifecycles = singleSpaAngular({
  unmountableStyles
})
```

internettrans:
Merging this since it documents the current behavior of single-spa-angular, but would like a response to my review comment, for my own understanding

versioned_docs/version-6.x/ecosystem-angular.md
arturovt: I'm concerned that exposing such functionality to users may lead to issues, as many people might struggle to import the SCSS file correctly. Since understanding how Webpack works can be challenging, I'm apprehensive about integrating it into the core.

PR #651, opened 2024-10-02
Document importMapUrl option in webpack-config-single-spa

internettrans:


internettrans:
See related https://github.com/single-spa/create-single-spa/pull/414

PR #664, opened 2025-01-26
Begin work on systemjs to esm migration guide

internettrans:
https://github.com/single-spa/create-single-spa/issues/433


ddavid93:
@jolyndenning so, can we finally use ESM SingleSpa on production?

internettrans:
Nothing is stopping you

PR #672, opened 2025-03-07
Add Tricentis as a new user 

damisparks:


src/data/users.js
```diff
@@ -188,6 +188,12 @@ const users = [
     infoLink: "https://epilot.cloud/",
     pinned: true,
   },
+  {
+    caption: "Tricentis",
+    image: "img/tricentis-logo.svg",
+    infoLink: "https://www.tricentis.com/",
+    pinned: true,
```
internettrans: No more pinned without donations

```suggestion
    pinned: false,
```

PR #681, opened 2025-08-01
Inconsistency in ecosystem-react.md: 

marapper:
The new ReactDOMClient parameter is skipped in the options list, which could be confusing (two examples above show it for both React versions).


internettrans (approved):

internettrans:
the docs will be rewritten for single-spa@7. CommonJS should never be encouraged.

